PR #1, opened 2022-01-12
draft add bazel for slsa

rbehjati:
Fixes #<issue_number_goes_here>

- [ ] Tests pass


tiziano88:
Nice! I'll take a closer look shortly! Curious to hear if @jul-sh thinks this is usable from Oak in this form, or are there things that need to be changed even before we can use it there?

common/common_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"path/filepath"
+	"testing"
+
+	cmp "github.com/google/go-cmp/cmp"
+	"github.com/project-oak/transparent-release/slsa"
+)
+
+var testdataPath = "../testdata/"
```
tiziano88, line 25: const?

common/common_test.go
```diff
@@ -0,0 +1,80 @@
...
+		Repo:                     "https://github.com/project-oak/oak",
+		CommitHash:               "a9ad36dede15386a6a9fa98d46aeede3205e2b29",
+		BuilderImage:             "gcr.io/oak-ci/oak@sha256:ff9eafaf9f64d6549039fd1e7c5a9163206d60495d0d232971c57fa5a52e7878",
+		Command:                  []string{"./scripts/runner", "build-functions-server"},
+		OutputPath:               "./oak_functions/loader/bin/oak_functions_loader",
+		ExpectedBinarySha256Hash: "9951f53ca22d9abdbbd664880586c4e2053087a5de891572458e84752ce1a8c1",
+	}
+
+	if cmp.Diff(got, want) != "" {
+		t.Errorf("invalid config: got %q, want %q", got, want)
```
tiziano88, line 78: Does %q actually print the struct? I would expect it prints a pointer in this case

slsa/BUILD
```diff
@@ -0,0 +1,35 @@
...
+
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+package(default_visibility = ["//:__subpackages__"])
+
+go_library(
+    name = "slsa",
+    srcs = ["slsa.go"],
+    importpath = "github.com/project-oak/transparent-release/slsa",
+    )
```
tiziano88, line 25: (nit) formatting

testdata/provenances/BUILD
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+#
+
+package(default_visibility = ["//:__subpackages__"])
+
+licenses(["notice"])
+
+exports_files([
+    "9951f53ca22d9abdbbd664880586c4e2053087a5de891572458e84752ce1a8c1.json",
+])
```
tiziano88, line 23: Is this a short for a filegroup?

tiziano88 (approved):

common/common_test.go
rbehjati: Yes. It does: 
```
    common_test.go:78: invalid config: got &{"https://github.com/project-oak/oak" "a9ad36dede15386a6a9fa98d46aeede3205e2b29" "gcr.io/oak-ci/oak@sha256:ff9eafaf9f64d6549039fd1e7c5a9163206d60495d0d232971c57fa5a52e7878" ["./scripts/runner" "build-functions-server"] "./oak_functions/loader/bin/oak_functions_loader" "9951f53ca22d9abdbbd664880586c4e2053087a5de891572458e84752ce1a8c1"}, want &{"https://github.com/project-oak/oak" "a9ad36dede15386a6a9fa98d46aeede3205e2b29-test" "gcr.io/oak-ci/oak@sha256:ff9eafaf9f64d6549039fd1e7c5a9163206d60495d0d232971c57fa5a52e7878" ["./scripts/runner" "build-functions-server"] "./oak_functions/loader/bin/oak_functions_loader" "9951f53ca22d9abdbbd664880586c4e2053087a5de891572458e84752ce1a8c1"}
```

Do you think it should be changed to `%v`?

common/common_test.go
rbehjati: Done.

slsa/BUILD
rbehjati: Done.

testdata/provenances/BUILD
rbehjati: What is a `filegroup`?

rbehjati (commented):
Thanks.

rbehjati:
I'll proceed and merge this PR, and will improve the code in future PRs. 
@jul-sh feel free to comment on this, or create an issue for anything that needs to be added or changed to make this usable in Oak. Of course this is still in very early stages, and we need to iterate on it :)

jul-sh:
Agreed, thanks for merging it @rbehjati, and glad to see that we got go + bazel to work together. :) 

RE your question @tiziano88, I'll +1 @rbehjati's take. Oak's release process isn't fully defined yet, and this tooling is also still WIP, so I cannot definitely answer that. I don't think we're painting ourselves into a corner. Feeling optimistic it'll work together nicely eventually.

rbehjati:
@jul-sh One idea that @tiziano88 and I were discussing was to automatically generate a simple config file similar to `teatdata/build.toml` for each PR in oak, and run `cmd/build/main.go` for it to generate a provenance file, and publish that in a branch on oak instead of the release process. All this can be done in a new workflow to replace the existing reproducibility workflow. 

Automatically generating the provenance files is not there yet, but it is easy to implement that for the v0 of our buildType spec.

jul-sh:
@rbehjati It's a good idea, I've been thinking about something along the same lines. The closer to CI/CD we can safely go, the better imo. :)

jul-sh:
As an aside to storing provenance, it might also be worth considering git notes, git's build in feature for post-commit commit metadata: https://git-scm.com/docs/git-notes

rbehjati:
Thanks @jul-sh. If you think it is a viable solution, feel free to write a proposal explaining how to use it. If the plan is to eventually use it for implementing the release inventory, or part of the release inventory, then the important thing is to be able to link the provenance files to Rekor log entries via endorsement files. This might mean that we'll end up having different ways of storing provenances and endorsement files. We should evaluate whether this is a good (e.g., easy to maintain) option.

PR #3, opened 2022-01-24
Add SLSA build type

jul-sh:
This adds a schema to describe a first version of our SLSA build type. It also adds an example of a full in-toto statement that utilises our statement. :)


amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+          ]
+        },
+        "outputPath": {
+          "type": "string",
+          "description": "The path, relative to the root of the git repository, where the artifact built by the command is expected to be found.",
+          "examples": [
+            "./oak_functions/loader/bin/oak_functions_loader"
+          ]
+        },
+        "expectedArtifactDigest": {
```
jul-sh, line 37: You'll notice that I slightly renamed this, to account for artifacts that may not be binaries. :)

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
+{
```
tiziano88, line 1: Could you add more details either here or to the commit message? I am unsure who / what is expected to consume this file at the moment, or is it just for documentation?

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+  "description": "Describes a SLSA buildType, as defined in https://slsa.dev/provenance/v0.2.",
+  "type": "object",
+  "required": [
+    "buildType",
+    "buildConfig",
+    "materials"
+  ],
+  "not": {
+    "required": [
+      "invocation"
```
tiziano88, line 13: Why is this not required?

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+        "outputPath": {
+          "type": "string",
+          "description": "The path, relative to the root of the git repository, where the artifact built by the command is expected to be found.",
+          "examples": [
+            "./oak_functions/loader/bin/oak_functions_loader"
+          ]
+        },
+        "expectedArtifactDigest": {
+          "type": "object",
+          "description": "Expected hash of the output artifact. Can be ommitted.",
```
tiziano88, line 39: What's the use of a build file that omits this?

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+    "materials": {
+      "type": "array",
+      "items": [
+        {
+          "title": "Docker Builder Image",
+          "type": "object",
+          "properties": {
+            "uri": {
+              "type": "string",
+              "description": "URI of a docker image on a registry",
```
tiziano88, line 71:
```suggestion
              "description": "URI of a Docker image on a registry",
```

amber-slsa-buildtype.json
tiziano88, line 1: To check my understanding @jul-sh @rbehjati : we will use this in the short term, and move to a standard SLSA provenance in the future? Or will we keep this?

jul-sh: Let's chat more in shared review, but the gist is that (afaik) there is no such thing as standard SLSA. SLSA is an umbrella, many things are SLSA :)

amber-slsa-buildtype.json
jul-sh: This goes back to the different layers of in-toto that I mentioned before. A SLSA provenance files usually do not seem include this information. It can be captured in the other layers of the in-toto attestation that SLSA is a part of https://github.com/in-toto/attestation/blob/main/spec/README.md#statement.

For convenience we felt having it in the SLSA file itself would be handy

amber-slsa-buildtype.json
tiziano88: Yes I get it, but GitHub and Google Cloud Build will probably go ahead and create their own (pseudo-)standard provenance files at some point, right? In which case we should probably support those rather than our own. Especially if those GitHub or Google provided ones are signed by GitHub or Google themselves (i.e. they would have greater authority that we can possibly infuse in our own builder, which no one else has any reason to trust).

rbehjati, line 1: Can you create a `schema/v0.2` directory and put this file in there? I believe we are going to have several versions of this file, so it is good to start with such a structure.

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+    "materials"
+  ],
+  "not": {
+    "required": [
+      "invocation"
+    ]
+  },
+  "properties": {
+    "buildType": {
+      "const": "https://github.com/project-oak/transparent-release/amber-slsa-buildtype.json"
```
rbehjati, line 18: I think it is important to have versioning here, so having a directory structure like the one I suggested is required.

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+        },
+        "outputPath": {
+          "type": "string",
+          "description": "The path, relative to the root of the git repository, where the artifact built by the command is expected to be found.",
+          "examples": [
+            "./oak_functions/loader/bin/oak_functions_loader"
+          ]
+        },
+        "expectedArtifactDigest": {
+          "type": "object",
```
rbehjati, line 38: Is it possible to clarify that it has to be DigestSet in https://github.com/in-toto/attestation/blob/main/spec/field_types.md#DigestSet? 

Do we have to define that object somewhere ourselves to be able to use json/toml validation?

amber-slsa-buildtype.json
rbehjati, line 39:
```suggestion
          "description": "Expected hash of the output artifact. Optional.",
```

nit: I think calling such fields optional is more conventional.

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+          "description": "Expected hash of the output artifact. Can be ommitted.",
+          "properties": {
+            "sha256": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "sha256"
+          ],
+          "additionalProperties": false,
```
rbehjati, line 48: Why is it false? I think it is fine for this version, but should we keep it for future versions too?

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+    "materials": {
+      "type": "array",
+      "items": [
+        {
+          "title": "Docker Builder Image",
+          "type": "object",
+          "properties": {
+            "uri": {
+              "type": "string",
+              "description": "URI of a Docker image on a registry",
```
rbehjati, line 71: Perhaps clarify here that the URI must specify the image digest.

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+              "examples": [
+                "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320",
+                "index.docker.io/library/ubuntu:20.04"
```
rbehjati, line 74: I think we don't want to accept these, for this version. 
(or we could, but the verification code should be updated to check the digest.)

rbehjati: Actually, I am not sure which of the versions we discussed this is now :) If we want to use this schema, the code and the examples must be updated. Can you create some issues to track the work that needs to be done?

amber-slsa-buildtype.json
```diff
@@ -0,0 +1,137 @@
...
+              "properties": {
+                "sha1": {
+                  "type": "string"
+                }
+              },
+              "required": [
+                "sha1"
+              ],
+              "additionalProperties": false,
+              "description": "The specific commit hash to checkout"
```
rbehjati, line 124: nit: Can you move the description up, right after the `type`, to follow the same order as in the rest of the file?

rbehjati:
Any ideas why `cla/google` is not happy? Have I misconfigured something?

amber-slsa-buildtype.json
rbehjati: Do such provenances currently exist? 

The code in this repository is a proof-of-concept. If we have provenances produced by Google Cloud Build or GitHub we can extend the tool in this repo with support for those formats.

tiziano88: That's fine by me, I was just curious (hence my other questions too). Even my other question about whether this schema gets actually used by anything is mostly out of curiosity, since my understanding is that nothing actually parses it mechanically, so it's effectively for human consumption, but I'm not sure, so I hope to learn more :)

tiziano88:
> Any ideas why `cla/google` is not happy? Have I misconfigured something?

I suspect what happened is that I suggested a fix in GitHub, and Juliette probably applied it by pressing the button that GitHub shows, but that creates a commit that is co-authored by me, using my personal email address, which is not covered by the CLA.

I suggest never to use the functionality to accept these fixes in the GitHub UI, and just author the fixes manually instead.

amber-slsa-buildtype.json
rbehjati: The plan is to use this schema both for the toml file and the json provenance file. If I understand correctly, it should be possible to automatically validate the toml and json files against this schema. But the example files in the repo not currently conform to it. That is why I have some questions about the versioning :) 

I am fine with keeping this version, but should as well update both the examples files and the code.

amber-slsa-buildtype.json
jul-sh, line 18: If we want to hash the buildType (see https://github.com/in-toto/attestation/issues/81 for context), we could ask clients to specify this as a [ni URL](https://www.rfc-editor.org/rfc/rfc6920.html). Using a format that is also a valid HTTP URL would be nice for human readability eg `ni://example.com/sha-256;f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk`. But it seems difficult to create a conformant HTTP link for inside a github repo.

jul-sh: Also worth noting that with hashes there's a bit of a general inconvenience that this schema file cannot include its own hash.

amber-slsa-buildtype.json
rbehjati: It should refer to the version of this schema. Chose whatever version you think makes sense :)

amber-slsa-buildtype.json
rbehjati: Actually, I don't see what is the benefit of a digest when the buildType is a schema like this. It is very easy to validate a json file against the schema. If the provider silently modifies the schema, it will easily be detected as the validation will fail. So, in case of a URI pointing to a json schema as the buildType, the consumer of the attestation does not have to trust the author of the schema. 

Let's not worry about the digests, at least for this version, and just use a URI.

jul-sh: Happy to skip validation for this version. :) I do think it still matters for JSON schema's. While the provider cannot change the structure, they can change what each key means. :)

amber-slsa-buildtype.json
rbehjati: Sounds good to me :)

amber-slsa-buildtype.json
rbehjati: We don't have `invocation` in this schema. So I guess it somehow refers to the one in https://slsa.dev/provenance/v0.2. Is that right? 

The in-toto statements must specify a `predicateType`, which in case of SLSA provenances is:
```
"predicateType": "https://slsa.dev/provenance/v0.2",
```

With this schema, can we still use `https://slsa.dev/provenance/v0.2` or do we have to use this schema?

jul-sh: You're correct that that's a reference to the SLSA schema. I'll add a comment to clarify that. We can still use their predicateType.

The reason this is omitted is that in SLSA, `invocation` and `buildConfig` hold the same information. We discussed this in the [Google document](https://docs.google.com/document/d/1xTEqPQxBG0iOSCbX1F79e1vx0zNelOoce00t6Z5-agU/edit?disco=AAAAUTPzORY) and agreed to use `buildConfig` only. :)

amber-slsa-buildtype.json
jul-sh: > Is it possible to clarify that it has to be DigestSet in https://github.com/in-toto/attestation/blob/main/spec/field_types.md#DigestSet?
Possible yes. Note that at the moment however this is a subset of the DigestSet, as it only accepts one type of hash. DigestSet allows arbitrary hashing functions. I'll add a comment to clarify. It is limited to `sha256` to allow our tooling to validate hashes. We can add more hash types in later versions. :) 

> Do we have to define that object somewhere ourselves to be able to use json/toml validation?
Yes

amber-slsa-buildtype.json
jul-sh: This goes back to my answer on https://github.com/project-oak/transparent-release/pull/3#discussion_r794467352. Our tooling can only realistically support a limited number of hashing functions.

We could allow other ones to be set as additional values, though we wouldn't be able to validate them. Perhaps that's fine though. :)

amber-slsa-buildtype.json
jul-sh: Can do! :) Yeah we'd need to check the digest.

amber-slsa-buildtype.json
jul-sh: Sure :)

amber-slsa-buildtype.json
rbehjati: Just to make sure I understand, this schema is not just a schema for the `buildType`. It is a **complete predicate specification** that conforms to the SLSA provenance spec (https://slsa.dev/provenance/v0.2#schema).

If so, I think it must include a `builder` too, since it is a [required field](https://slsa.dev/provenance/v0.2#fields) in the SLSA provenance.

amber-slsa-buildtype.json
rbehjati: Yes. I think it is ok for the schema to allow other ones, as long as it requires `sha256`. We only validate `sha256` for now.

amber-slsa-buildtype.json
jul-sh: In my understanding this schema would merely lint that a SLSA file conforms to our build type. It does not actually check if it is a valid SLSA file. I went with this route so we wouldn't have to duplicate the SLSA schema definition, but happy to reconsider this approach.

rbehjati: Do you have an example showing how the final in-toto statement would look like? The confusing part for me is that this schema mentions parts of the SLSA provenance (e.g., materials, buildConfig, and invocation) but not all of it. It is also unclear to me how we could use this schema for validating in-toto statements, with SLSA provenances and this build type.

jul-sh: Can do! There's an example of how a provenance in-toto statement would look like in the in-toto repo: https://github.com/in-toto/attestation#provenance-example

I also created an example that uses our SLSA build type:

```
{
  "_type": "https://in-toto.io/Statement/v0.1",
  "subject": [
    {
      "name": "oak_functions_loader",
      "digest": {
        "sha256": "d4d5899a3868fbb6ae1856c3e55a32ce35913de3956d1973caccd37bd0174fa2"
      }
    }
  ],
  "predicateType": "https://slsa.dev/provenance/v0.2",
  "predicate": {
    "builder": {
      "id": "https://github.com/project-oak/transparent-release"
    },
    "buildType": "https://github.com/project-oak/transparent-release/amber-slsa-buildtype/v1.json",
    "buildConfig": {
      "command": "./scripts/runner build-functions-server",
      "outputPath": "./oak_functions/loader/bin/oak_functions_loader"
    },
    "materials": [
      {
        "uri": "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320",
        "digest": {
          "sha256": "15dc16c42a4ac9ed77f337a4a3065a63e444c29c18c8cf69d6a6b4ae678dca5c"
        }
      },
      {
        "uri": "https://github.com/project-oak/oak",
        "digest": {
          "sha1": "0f2189703c57845e09d8ab89164a4041c0af0a62"
        }
      }
    ]
  }
}

```

If you'd like I could include an example file in this PR :)

jul-sh: Update: added an example to the PR :)

jul-sh: Also happy to hear your opinion if we should change the scope of the schema to validate that the statement is a valid SLSA file as well. 

My first intuition was no (to keep the scope small), but I realized that it would only require a few additional keys:
`_type`, `subject`, `predicate.builder.id`.

schema/v1-example-statement.json
```diff
@@ -0,0 +1,36 @@
...
+    "buildType": "https://github.com/project-oak/transparent-release/amber-slsa-buildtype/v1.json",
+    "buildConfig": {
+      "command": "./scripts/runner build-functions-server",
+      "outputPath": "./oak_functions/loader/bin/oak_functions_loader"
+    },
+    "materials": [
+      {
+        "uri": "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320",
+        "digest": {
+          "sha256": "15dc16c42a4ac9ed77f337a4a3065a63e444c29c18c8cf69d6a6b4ae678dca5c"
```
tiziano88, line 25: Why is this hash different from the one in the digest of the URI? Maybe it's obvious to you or others who know more about Docker, but could you add a comment anyways?

schema/v1-example-statement.json
```diff
@@ -0,0 +1,36 @@
...
+        }
+      },
+      {
+        "uri": "https://github.com/project-oak/oak",
```
tiziano88, line 29: Is this actually how the commit is referred to in the build? That seems to be the definition of this field in https://slsa.dev/provenance/v0.2 (which I find somewhat confusing myself, since it seems to assume that the URI is stable / immutable (which here is not, obviously). Or is there a URI format for git commits that includes the hash of the commit, similar to how the Docker URI includes the digest of the image. Anyways please don't block on this, it may be something that we can bring up with the SLSA people.

Also unfortunately my usual suggestion of using https://github.com/google/ent does not apply here, since git still uses sha1, and Ent only supports sha256 (so far) :sweat:

tiziano88 (approved):

schema/v1-example-statement.json
```diff
@@ -0,0 +1,36 @@
...
+        "sha256": "d4d5899a3868fbb6ae1856c3e55a32ce35913de3956d1973caccd37bd0174fa2"
+      }
+    }
+  ],
+  "predicateType": "https://slsa.dev/provenance/v0.2",
+  "predicate": {
+    "builder": {
+      "id": "https://github.com/project-oak/transparent-release"
+    },
+    "buildType": "https://github.com/project-oak/transparent-release/amber-slsa-buildtype/v1.json",
```
tiziano88, line 16: I take it that the assumption here is that the docker image is mounted in a specific way, and that the git commit is cloned at a specific folder, and this command is executed in a specific manner and by a specific user? And in principle, if we change any of these things we would have to bump the version of the build type, since it would be incompatible with the previous version, right? Not to worry about that for now, but that's also what I meant when I said that there may probably be some SLSA provenance format that describes a generic way of putting together a docker image with a git commit and run a command (either now or in the future).

tiziano88 (approved):

schema/v1-example-statement.json
jul-sh: At the moment it seems this is the way, yeah. I also wondered about a URI format for cloning specific git commits, but that doesn't seem to be a thing. 

There is an in-toto issue about differentiating between different "kinds" of digests that might improve this: https://github.com/in-toto/attestation/issues/28

As an aside, you probs know this already, but git does plan to move to sha256, but afaik it's taking a while as it's part of larger refactor to make git hash function agnostic.

schema/v1-example-statement.json
jul-sh: Nah, the difference in hash is literally just a mistake a made when assembling the example. 😅 Thx for catching, will fix on Mon

schema/v1-example-statement.json
jul-sh: In the future there likely will be some standard for that, yes. There isn't one yet, we're all still in the initial experimental phase.

Ideally all the actually relevant "specific ways, manners, etc" relevant for a build should be captured in this build type. But you're right that it cannot actually be truly comprehensive and defacto there are and likely always will be implicit defaults. We'll bump versions as needed with changes to those. Ideally that iterative process will eventually yield the the stable, "generic way" you alluded to.

schema/v1.json
```diff
@@ -0,0 +1,113 @@
...
+    "predicate": {
+      "type": "object",
+      "required": [
+        "buildType",
+        "buildConfig",
+        "materials"
+      ],
+      "properties": {
+        "buildType": {
+          "const": "https://github.com/project-oak/transparent-release/amber-slsa-buildtype/v1.json"
```
rbehjati, line 19: This path is not valid. It should also be fixed in the example. 

Also, having a `v1.json` directly under `schema` does not look like a good file structure to me. I will soon add a schema for endorsements, which will also go somewhere in the `schema`  folder.  We need a better file structure. 

I can think of two options:

```
transparent-release
|-- schemas
|    |-- v_i
|    |    |-- amber_slsa_buildtype.json
|    |    |-- amber_endorsement.json
|    |    |-- some_example_file.json
|    |-- v_j
|-- ...
```

or 

```
transparent-release
|-- schemas
|    |-- amber_slsa_buildtype
|    |    |-- some_example_file.json
|    |    |-- v_i.json
|    |    |-- v_j.json
|    |-- amber_endorsement
|    |    |-- v_i.json
|-- ...
```

I don't have a strong preference between these two options, but the latter allows for the endorsement and the build-type to evolve separately, which might be what we want. The former, makes more sense if we look at the set of the schemas for the transparent-release as a whole that we'd publish and release together.

amber-slsa-buildtype.json
rbehjati: Thanks. Let's add some code that validates the example against the schema (in a separate PR). We can add anything that is missing afterwards :)

schema/v1.json
jul-sh: Good point on both the link and the structure. I like the second one. I think the ability to update them separately is good to have at this point. :)

schema/amber-slsa-buildtype/v1-example-statement.json
```diff
@@ -0,0 +1,36 @@
...
+        "sha256": "d4d5899a3868fbb6ae1856c3e55a32ce35913de3956d1973caccd37bd0174fa2"
+      }
+    }
+  ],
+  "predicateType": "https://slsa.dev/provenance/v0.2",
+  "predicate": {
+    "builder": {
+      "id": "https://github.com/project-oak/transparent-release"
+    },
+    "buildType": "https://github.com/project-oak/transparent-release/schemas/amber-slsa-buildtype/v1.json",
```
rbehjati, line 16:
```suggestion
    "buildType": "https://github.com/project-oak/transparent-release/schema/amber-slsa-buildtype/v1.json",
```

schema/amber-slsa-buildtype/v1.json
```diff
@@ -0,0 +1,113 @@
...
+    "predicate": {
+      "type": "object",
+      "required": [
+        "buildType",
+        "buildConfig",
+        "materials"
+      ],
+      "properties": {
+        "buildType": {
+          "const": "https://github.com/project-oak/transparent-release/schemas/amber-slsa-buildtype/v1.json"
```
rbehjati, line 19:
```suggestion
          "const": "https://github.com/project-oak/transparent-release/schema/amber-slsa-buildtype/v1.json"
```

rbehjati (approved):
Nice. Thanks :)

PR #9, opened 2022-02-01
Make command a string

jul-sh:
It is parsed as a string, hence we'd like to change it. The SLSA amber build type uses a string as well.

This is the first in a series of small PRs to implement #8.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -42,7 +42,7 @@ type BuildConfig struct {
 	// URI identifying the Docker image to use for building the binary.
 	BuilderImage string `toml:"builder_image"`
 	// List of commands to pass to the docker run command.
-	Command []string `toml:"command"`
+	Command string `toml:"command"`
 	// The path, relative to the root of the git repository, where the binary
 	// built by the `docker run` command is expected to be found.
 	OutputPath string `toml:"output_path"`
@@ -129,7 +129,7 @@ func (b *BuildConfig) Build() error {
 	args = append(args, "run")
 	args = append(args, defaultDockerRunFlags...)
 	args = append(args, b.BuilderImage)
-	args = append(args, b.Command...)
+	args = append(args, b.Command)
 	cmd := exec.Command("docker", args...)
 
 	stderr, err := cmd.StderrPipe()
--- a/common/common_test.go
+++ b/common/common_test.go
@@ -69,7 +69,7 @@ func checkBuildConfig(got *BuildConfig, t *testing.T) {
 		Repo:                     "https://github.com/project-oak/oak",
 		CommitHash:               "0f2189703c57845e09d8ab89164a4041c0af0a62",
 		BuilderImage:             "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320",
-		Command:                  []string{"./scripts/runner", "build-functions-server"},
+		Command:                  "./scripts/runner build-functions-server",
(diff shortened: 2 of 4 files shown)


rbehjati (approved):
Nice. Thanks :)

tiziano88 (commented):
I'm not convinced about this, how will it handle multiple arguments and escaping? I think we should keep it as an array, similar to how it is in Docker, but also generally for `execve`-like invocations

rbehjati:
> I'm not convinced about this, how will it handle multiple arguments and escaping? I think we should keep it as an array, similar to how it is in Docker, but also generally for `execve`-like invocations

Thanks @tiziano88. I just noticed that the verify functionality no longer works, now that the command is a string. 

@jul-sh Let's try and add more tests and CI steps, before making any other changes. 

I have created #/12 that adds a test for verify. It currently fails on GitHub, and I cannot find a way to show the logs from bazel in the console. Do you know how this can be fixed @tiziano88?

tiziano88:
My question still stands, regardless of whether the build is fixed: if command is a string, how are multiple arguments provided? Is this a limitation of SLSA?

@jul-sh

jul-sh:
@tiziano88 Not a limitation of SLSA. Multiple commands could be joined with `&&`. We can also use an array to provide multiple commands. 

It's worth noting though that this is orthogonal to this change. The array this replaced is not in fact for multiple commands. It denotes different segment which are parsed executed as a single command.

tiziano88:
I did not mean multiple commands, just multiple arguments to a sngle command. I'm still not sure how that would work with command being a single string?

tiziano88:
(Also, `&&` is a shell construct, so I would not expect it to work to join multiple commands, anyways that was not my question, and I don't think it will be an issue for us, but let me know if you think we need something like that too)

jul-sh:
Okay, thanks for clarifying. It would work the same way you type a command with multiple arguments into the shell prompt? 
eg `git commit --date="10 day ago" -m "Your commit message"`

tiziano88:
In that case the shell is doing a lot of parsing and escaping to produce the actual arguments to pass to execve or equivalent call. 

i.e. `git commit --date="10 day ago" -m "Your commit message"` results in the shell parsing the following arguments:

- `git`
- `commit`
- `--date=10 day ago`
- `-m`
- `Your commit message`

And then invoking `git` with those commands. Then in any programming language, these arguments are available in argv or similar parameter to main.

But we don't have a shell around, and even if we did, I don't think it's helpful to use it to just undo what we are doing by concatenating the arguments together in order to pass them as a single string. Also the rules for escaping quotes, double quotes, etc would depend on the shell, so we cannot just replicate it by splitting on whitespaces, we would really have to reimplement a tiny shell language parser, for which I see no particularly good reason.

In fact, this is exactly the issue that ended up happening in https://github.com/project-oak/transparent-release/pull/12 , since building a command line invocation requires passing the arguments as an array (in Go but also in any other language, and also in Docker).

tiziano88:
Sure, glad it helped! To be honest I might have gotten confused by that very array in the past too, since it gives the idea of being a list of commands rather than a single command (as you noted). It may be worth clarifying in some comments what the semantics actually is. Perhaps we can also reference https://docs.docker.com/engine/reference/builder/#cmd or https://man7.org/linux/man-pages/man3/exec.3.html if it's still ambiguous, and provide some examples in the comments so that we (or users of this package) don't do this mistake in the future.

tiziano88:
I think the main take away is to be careful when using command vs arguments; if we do that consistently, then it should be fine.

rbehjati:
Thanks. I've updated the misleading comment in #12.

PR #10, opened 2022-02-01
Remove existing SLSA parsing in preparation for #8

jul-sh:
Remove existing SLSA parsing in preparation for #8, my next PR will add the new implementation. Separating them to keep the diff small, and PRs easier to review


verify/verify.go
```diff
@@ -1,65 +0,0 @@
-// Copyright 2022 The Project Oak Authors
```
rbehjati, line 1: Why is this deleted? We need the verification logic.

testdata/provenances/15dc16c42a4ac9ed77f337a4a3065a63e444c29c18c8cf69d6a6b4ae678dca5c.json
```diff
@@ -1,52 +0,0 @@
-{
```
rbehjati, line 1: Why is this deleted? I think it is good to have an example SLSA provenance file. I think it is better to update it according to our new format, and add logic that verifies it.

slsa/slsa.go
```diff
@@ -1,82 +0,0 @@
...
-// Parameters represents invocation parameters in a SLSA provenance file.
-type Parameters struct {
-	Repository     string
-	CommitHash     string `json:"commit_hash"`
-	BuilderImage   string `json:"builder_image"`
-	Command        string
-}
-
-// ParseProvenanceFile parses a JSON file in the given path into a Provenance object.
-func ParseProvenanceFile(path string) (*Provenance, error) {
```
rbehjati, line 70: We still need to parse provenance files, using the schema. Right? How is that done?

rbehjati (commented):
I am not sure why `slsa` and `verify` packages are removed without being replaced with anything. This repo is not only about building the binaries and generating the provenances. The main part is the verification of a provenance.

jul-sh:
@rbehjati Please see the PR description which answers your questions.

> Remove existing SLSA parsing in preparation for #8, my next PR will add the new implementation. Separating them to keep the diff small, and PRs easier to review

I did this in the interest of small PRs, as I've discussed with you on. Closing this to instead open a larger PR 🤷 that includes the new implementation.

rbehjati:
Thanks. Small PRs should not just remove functionality in my opinion :)

rbehjati:
Without knowing your plan for the changes, I'd suggest starting with #6:
- add functionality for validating a json against the schema, perhaps in the `slsa` package
- add a simple test that verifies the existing example json (in `schema`) against the schema
- use the same test in the GitHub CI

Then in a separate PR update `verify` to use the same code for parsing a json file into an object. 

Then in a separate PR proceed to #8, and do the refactoring you have in mind. Or by TOML format in #8 do you mean the current BuildConfig type?

jul-sh:
@rbehjati Thank you for the additional context, that is helpful. As mentioned in our 1-1 chat stream I've implemented schema validation and parsing/verifying of Amber SLSA files in a small vanilla go CLI (https://github.com/jul-sh/transparent-release/tree/vanilla-slsa in case you're curious). I'm happy to port the functionality over in small PRs to make it easy to review. 

If you'd like I can start with just porting over the functionality to validate the amber schema to begin with, putting that functionality into the slsa module. I'll make my next PR that.

jul-sh:
With regards to porting over the functionality to parse the build type and re-run the build: my understanding following our meeting discussing the build type on Thursday was that we would like to deprecate support for the verifying builds described in a TOML file (mirrored the `BuildConfig` struct) and instead use amber SLSA files only. 

Hence the implementation that I wrote (and would to port over in a future PRs) removes that functionality. We can talk about the details later, but I did want to confirm whether removing support for TOML files is indeed (still) the plan? :)

PR #12, opened 2022-02-03
Test verify

rbehjati:
Changes:
- Adds a test for verify
- Updates a misleading comment in the definition of `BuildConfig`
- Reverts #9 for making command a string

Checklist:
- [x] Tests pass

### DIFF
--- a/testdata/build.toml
+++ b/testdata/build.toml
@@ -1,6 +1,6 @@
 repo = "https://github.com/project-oak/oak"
 commit_hash = "0f2189703c57845e09d8ab89164a4041c0af0a62"
 builder_image = "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320"
-command = "./scripts/runner build-functions-server"
+command = ["./scripts/runner", "build-functions-server"]
 output_path = "./oak_functions/loader/bin/oak_functions_loader"
 expected_binary_sha256_hash = "15dc16c42a4ac9ed77f337a4a3065a63e444c29c18c8cf69d6a6b4ae678dca5c"
--- a/verify/BUILD
+++ b/verify/BUILD
@@ -30,8 +30,7 @@ go_library(
 
 go_test(
     name = "verify_test",
-    size = "small",
-    timeout = "long",
+    size = "large",
     srcs = ["verify_test.go"],
     data = [
         "//testdata/provenances:15dc16c42a4ac9ed77f337a4a3065a63e444c29c18c8cf69d6a6b4ae678dca5c.json",
--- a/verify/BUILD
+++ b/verify/BUILD
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 package(default_visibility = ["//:__subpackages__"])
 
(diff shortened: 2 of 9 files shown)


testdata/build.toml
```diff
@@ -1,6 +1,6 @@
 repo = "https://github.com/project-oak/oak"
 commit_hash = "0f2189703c57845e09d8ab89164a4041c0af0a62"
 builder_image = "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320"
-command = "./scripts/runner build-functions-server"
+command = ["./scripts/runner", "build-functions-server"]
```
jul-sh, line 4: This seems to reverse the change done in https://github.com/project-oak/transparent-release/pull/9. Are we sure we want to do that?

rbehjati: Yes. I had to revert the change in #9 to make the `verify` functionality work. 

It is important to pass an array in `args = append(args, b.Command...)` instead of a string, otherwise the command execution in the following line fails: 
https://github.com/project-oak/transparent-release/blob/b5c32a38184c44821e2d5abe71022e5fcf2b2ba5/common/common.go#L133

I could keep the input command as a single string, and split that into an array and use the array for building `args` right before executing the command. But I am not sure how that addresses the concern that @tiziano88 raised on #9. 

So let's get back to #9 after we have this test in CI :)

rbehjati: But feel free to create a PR to fix the array part, keeping the command as a single string. Then I'll rebase this on your change, instead of reverting #9 :)

jul-sh (approved):

verify/BUILD
```diff
@@ -27,3 +27,14 @@ go_library(
     ],
         importpath = "github.com/project-oak/transparent-release/verify",
 )
+
+go_test(
+    name = "verify_test",
+    size = "small",
```
jul-sh, line 33: Should we change this to larger size? This tests clones the Oak repo and runs a full build of it, which takes about 9min in CI.

verify/verify_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+)
+
+func TestVerifyProvenanceFile(t *testing.T) {
+	path := "../testdata/provenances/15dc16c42a4ac9ed77f337a4a3065a63e444c29c18c8cf69d6a6b4ae678dca5c.json"
+
+	if err := Verify(path, ""); err != nil {
+		t.Fatalf("couldn't verify the provenance file: %v", err)
+	}
```
jul-sh, line 26: It would be nice if this test could receive and check the actual hash of the build being perform. At the present it implicitly relies on `Verify` checking the hash.

rbehjati: This test is checking the functionality of `Verify`. I think the scenario you have described can be another, finer-grained test to be added to `common`, as part of #13.

rbehjati (commented):
Thanks for the review :)

verify/BUILD
jul-sh: I would add that I wrote an implementation this test yesterday using a simple mock build file (uses a 5MB alpine linux docker to produce an output "artifact" with a 5L bash script). Verifies the same thing, but runs a lot faster. Maybe we want to switch to something like that?

rbehjati: > I wrote an implementation of this type of test yesterday

I am not sure I follow. What do you mean by "this type of test"?

jul-sh: An integration test. In this specific case that means it runs an actual build and verifies the hash. The same thing as this, but it uses a simplified provenance file that executes faster.

jul-sh: We can have both too btw, it seems handy to run a test of verifying an actual Oak provenance file (since that is our current customer). But probably we want to separate out that test from the other unit tests, since it takes a long time to run. :)

rbehjati: The long running test on the CI should not be a problem. When testing locally, it would be nice to filter them out. Not sure what is the best/easiest way to do this with Bazel.

> The same thing as this, but it uses a provenance file that describes a simplified (or mock build) so it executes a lot faster.

That is a good idea. I think it could be a new test in `common_test.go` with a simple BuildConfing that could potentially use the same docker image from Oak, but run a different/simpler command. Is that what you are thinking?

Issue #20, opened 2022-02-08
What is the right SLSA Schema Format?


jul-sh:
We presently use a JSON schema to specify the shape of the expected SLSA buildType. While I went with a JSON schema, this was done initially just to have *some* clearly defined format. It doesn't have to be JSON schema specifically and there's a few pros & cons to them:

Pros:
- Machine readable, hence clearly defined
- Existing tools for both validation of schemas and for generating example data off them
- Open standard

Cons:
- They're like actually quite hard to read for humans
- We can't actually use them to parse JSON files, only to validate. Hence we require go structs which essentially duplicate the schema (in less detail). See PR #16 for deets.

Thoughts? Ideas? :) cc @rbehjati @tiziano88 

[This issue is related to #15, probably a superset of it].

jul-sh:
@tiziano88 suggested just using go structs for now, and I'm like personally rather partial to that idea. Imo the less duplication the better. If the go structs that represent data internally could also validate it that'd be great. 

The downside is that they don't seem to be quite as specific. Eg unmarshaling JSON silently fails if eg keys are not defined. We can write logic to catch it, but this does come at the expense of human readability. 

Imo this approach might be a good fit anyways though, at least initially. As we're still at the beginning of this project it's reasonable to think that we'll make schema changes a bit more frequently at this point. Only maintaining a single representation makes that a lot easier.

Once our schema is more stable we can always offer schema's of different formats for improved human readability.

tiziano88:
Since we need a Go struct *at some point* anyways, I indeed suggest starting from just that. We can document additional constraints (e.g. number of elements in a slice, ordering, allowed empty fields) as comments on the struct fields for now, and implement a validator / parser method to enforce those constraints. Unfortunately Go makes the usual [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) approach not very easy to implement because of its poor type system, but we have to work with that (and we can improve things over time). 

Once we have this struct, we can write most unit tests based on it (we don't need to deserialize a JSON / TOML file each time), and we can even decide to switch between JSON / TOML separately. IMO the validation logic should not be completely tied to JSON schema, it should be a validation of a Go struct (so that it can also be tested in our own code).

[...]

PR #21, opened 2022-02-08
Endorsement v0.1 schemas and example

rbehjati:
Schema and an example for Endorsement v0.1 predicate. An endorsement file is an in-toto statement that specifies a validity time-range that is required for implementing passive revocation when releasing, signing, and publishing binaries in a transparency log.

### DIFF
--- a/schema/amber-endorsement/v0.1/example.json
+++ b/schema/amber-endorsement/v0.1/example.json
@@ -1,0 +1,18 @@
+{
+    "_type": "https://in-toto.io/Statement/v0.1",
+    "subject": [
+      {
+        "name": "oak_functions_loader-0f2189703c57845e09d8ab89164a4041c0af0a62",
+        "digest": {
+          "sha256": "15dc16c42a4ac9ed77f337a4a3065a63e444c29c18c8cf69d6a6b4ae678dca5c"
+        }
+      }
+    ],
+    "predicateType": "https://github.com/project-oak/transparent-release/schema/amber-endorsement/v0.1/predicate.json",
+    "predicate": {
+      "validityPeriod": {
+        "releaseTime": "2022-02-01T10:20:50.32Z",
+        "expiryTime": "2022-03-01T10:20:50.32Z"
+      }
+    }
+  }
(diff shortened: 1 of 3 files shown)


tiziano88:
See my comments https://github.com/project-oak/transparent-release/issues/20#issuecomment-1032615988 and https://github.com/project-oak/transparent-release/issues/20#issuecomment-1032664429

There I was suggesting to first iterate on the Go structs (with appropriate comments) and implement the validation logic in Go, instead of using JSON schemas, since we would need Go structs anyways, and it's unclear who would even use the JSON schemas and for what yet.

WDYT? cc @jul-sh

jul-sh (approved):
Nice! We should also add some tests to validate that examples follow the schema. After merging #16 this should be relatively straightforward to do, as #16 does that for the other schema.

jul-sh:
@tiziano88 I would suggest keeping that discussion to #20 :). I'm not opposed to that proposal, but I also think it falls into the category of "refactoring", as we already use JSON schemas. Ergo it wouldn't be blocking for this PR. 

On a high level I'd also like to make sure we don't spend too much time bike shedding on formats, imo the codebase is small at the moment, the cost of changing them isn't that high.

tiziano88 (approved):
Cool, SGTM

rbehjati:
Thanks :) 

I agree with your concerns about using JSON schemas @tiziano88, and I agree with @jul-sh that we can refactor the code later according to the discussions in #20 :)

Actually, my main purpose with this PR is to provide some sort of documentation, and an example endorsement file. I'll adde a test as @jul-sh suggested in a separate PR. 

I also discovered that in-toto is not very strict about the `predicateType`. In fact, `https://in-toto.io/Statement/v0.1`, which is used as the value of `_type`, returns 404! So we don't need to maintain a JSON schema to have a predicate type. Perhaps, in future, we can replace the JSON schema with a README (that can be generated from the JSON schema using `jsonschema2md`), and rely on `go` structs for verification. 

Also, we don't yet have a concrete plan for consuming endorsement files. I think the first tool that will consume  endorsement file is the authorization logic wrapper (see https://github.com/google-research/raksha/pull/289#discussion_r801887166). We have also discussed the possibility of implementing an end-to-end tool with Rekor and claim-store integration, which I think we'll start working on in the summer. These are the tools that will actually benefit from the `go` structs and the validation logic.

PR #22, opened 2022-02-09
Test for validating the endorsement example

rbehjati:
Adds a test as suggested in https://github.com/project-oak/transparent-release/pull/21#pullrequestreview-877325031.

### DIFF
--- a/schema/amber-endorsement/v0.1/endorsement_test.go
+++ b/schema/amber-endorsement/v0.1/endorsement_test.go
@@ -1,0 +1,65 @@
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema
+
+import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


rbehjati:
@jul-sh Following your suggestion in https://github.com/project-oak/transparent-release/pull/21#pullrequestreview-877325031, I've added this test. I am not sure if this is the right place for the test though. 

I think we could rename the `slsa` package to `amber`, and have separate go files in it for provenance and endorsement, and their corresponding tests. WDYT?

jul-sh:
@rbehjati Thanks for adding that! 

> I think we could rename the slsa package to amber, and have separate go files in it for provenance and endorsement, and their corresponding tests. WDYT?

Yeah, why not, works for me. :). Deferring to you in this, I'm not terribly opinionated about what is the best place for the tests to live.

schema/amber-endorsement/v0.1/endorsement_test.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func TestExampleAmberEndorsement(t *testing.T) {
+	schemaPath := "statement.json"
+	examplePath := "example.json"
+
+	schemaLoader, err := loadJSON(schemaPath)
+	if err != nil {
+		t.Fatalf("Couldn't load schema file %v: %v", schemaPath, err)
+	}
```
jul-sh, line 34: I do low key wonder what the error ends up looking, since `loadJSON` and this add a description :)

jul-sh (approved):
LGTM :]

schema/amber-endorsement/v0.1/endorsement_test.go
rbehjati: Here is an example:
```
Couldn't load schema file statements.json: couldn't read json file statements.json: open statements.json: no such file or directory
```

I think, at least internally, it is conventional for each nested error to add a context-specific text.

PR #32, opened 2022-03-23
Souffle Binary with Static C++ Runtime Libs

aferr:
This PR fixes the CI regression by using a souffle binary that statically links the C/C++ runtime libraries. See also https://github.com/google-research/raksha/pull/484

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,10 @@ jobs:
           mkdir -p "${GITHUB_WORKSPACE}/bin/"
           mv bazelisk-linux-amd64 "${GITHUB_WORKSPACE}/bin/bazel"
           chmod +x "${GITHUB_WORKSPACE}/bin/bazel"
-      - name: Install mcpp #mcpp is a dependency of Souffle
-        run: sudo apt-get install -y mcpp 
+      - name: Install souffle deps #install dependencies of Souffle
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y mcpp libstdc++6 libc6-dev libc6
       - name: Test
         run: |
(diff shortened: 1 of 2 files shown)


tiziano88 (approved):
Thanks for the quick fix!

rbehjati:
I am still getting the same error when I run `bazel build //...` locally. Is a specific version of `bazel` needed for this to work?

tiziano88:
> I am still getting the same error when I run `bazel build //...` locally. Is a specific version of `bazel` needed for this to work?

It seems to work in CI though, or is it not being run there yet?

aferr:
> I am still getting the same error when I run `bazel build //...` locally. Is a specific version of `bazel` needed for this to work?

What sort of machine are you running this on? It will only work on linux machines. I can run it on a glinux cloudtop instance.

Sadly I think pulling in binaries is never going to be 100% self-contained.

rbehjati:
I am running it on my gLinux. 

Yes, it seems to run on CI. So it is very strange that it is failing for me locally.

aferr:
> > It seems to work in CI though, or is it not being run there yet?
> 
> Yes, it is running in CI. The authorization logic test is covered in build-all and AFAICT github is not taking crash faults today :)

Yikes... I meant to reply but edited your comment. Sorry about that.

rbehjati:
The error is `failed to locate mcpp pre-processor`. Is that familiar?

rbehjati:
I think it would nice to have development guidelines for this, but I noticed that we don't yet have a `docs/DEVELOPMENT.md` file. Would be nice to add that in a separate PR.

rbehjati (approved):

PR #36, opened 2022-03-25
Verifier wrapper

aferr:
This is a wrapper that takes an application name and produces the verification policy similar to the [Verification Policy Generator here](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/transparent_release_prototype/transparent_release.auth_logic)

### DIFF
--- a/experimental/auth-logic/verifier_wrapper.go
+++ b/experimental/auth-logic/verifier_wrapper.go
@@ -23,28 +23,31 @@ import (
 
 type verifierWrapper struct{ appName string }
 
+// This produces the policy code for checking if a binary can act as an
+// application by aggregating all the evidence from the other parties.
 func (v verifierWrapper) EmitStatement() UnattributedStatement {
-	endorsementPrincipal := fmt.Sprintf("\"%v::EndorsementFile\"", v.appName)
-	provenancePrincipal := fmt.Sprintf("\"%v::Provenance\"", v.appName)
-	binaryPrincipal := fmt.Sprintf("\"%v::Binary\"", v.appName)
-	appPrincipal := fmt.Sprintf("\"%v\"", v.appName)
+	// TODO(#39) consider using a [template](https://pkg.go.dev/text/template) to implement this.
+	endorsementPrincipal := fmt.Sprintf(`"%s::EndorsementFile"`, v.appName)
+	provenancePrincipal := fmt.Sprintf(`"%s::Provenance"`, v.appName)
+	binaryPrincipal := fmt.Sprintf(`"%s::Binary"`, v.appName)
+	appPrincipal := fmt.Sprintf(`"%s"`, v.appName)
 
-  // The verifier needs to import expected hashes from both the endorsement
-  // and provenance files. If we use the same predicate to represent both of 
-  // these statements in this SecPal-based syntax for authorization logic, we 
-  // will lose track of who originated each statement. For example, if we just 
-  // used `Binary has_expected_hash(<hash>)` and the verifier delegates this 
-  // predicate to both the endorsement file and the provenance file, we cannot 
-  // write a policy that looks for the same predicate from both. To work around
-  // this we had a second argument to the predicate to track the original
-  // speaker.
+	// The verifier needs to import expected hashes from both the endorsement
+	// and provenance files. If we use the same predicate to represent both of
+	// these statements in this SecPal-based syntax for authorization logic, we
+	// will lose track of who originated each statement. For example, if we just
+	// used `Binary has_expected_hash(<hash>)` and the verifier delegates this
+	// predicate to both the endorsement file and the provenance file, we cannot
+	// write a policy that looks for the same predicate from both. To work around
+	// this we had a second argument to the predicate to track the original
+	// speaker.
 
 	endorsementHashDelegation :=
-		fmt.Sprintf("%v canSay %v has_expected_hash_from(any_hash, %v).\n",
+		fmt.Sprintf("%s canSay %s has_expected_hash_from(any_hash, %s).\n",
 			endorsementPrincipal, binaryPrincipal, endorsementPrincipal)
 
 	provenanceHashDelegation :=
-		fmt.Sprintf("%v canSay %v has_expected_hash_from(any_hash, %v).\n",
+		fmt.Sprintf("%s canSay %s has_expected_hash_from(any_hash, %s).\n",
 			provenancePrincipal, binaryPrincipal, provenancePrincipal)
 
 	provenanceDelegation :=
@@ -56,15 +59,13 @@ func (v verifierWrapper) EmitStatement() UnattributedStatement {
 	rekorLogCheckDelegation :=
 		"\"RekorLogCheck\" canSay some_object canActAs \"ValidRekorEntry\".\n"
 
-	tab := "    "
-
 	binaryIdentificationRule :=
 		binaryPrincipal + " canActas " + appPrincipal + " :-\n" +
-			tab + binaryPrincipal + " hasProvenance(" + provenancePrincipal + "),\n" +
-			tab + endorsementPrincipal + " canActAs \"ValidRekorEntry\",\n" +
-			tab + binaryPrincipal + " has_expected_hash_from(binary_hash, " + endorsementPrincipal + "),\n" +
-			tab + binaryPrincipal + " has_expected_hash_from(binary_hash, " + provenancePrincipal + "),\n" +
-			tab + binaryPrincipal + " has_measured_hash(binary_hash).\n"
+			"\t" + binaryPrincipal + " hasProvenance(" + provenancePrincipal + "),\n" +
+			"\t" + endorsementPrincipal + " canActAs \"ValidRekorEntry\",\n" +
+			"\t" + binaryPrincipal + " has_expected_hash_from(binary_hash, " + endorsementPrincipal + "),\n" +
+			"\t" + binaryPrincipal + " has_expected_hash_from(binary_hash, " + provenancePrincipal + "),\n" +
+			"\t" + binaryPrincipal + " has_measured_hash(binary_hash).\n"
 
 	return UnattributedStatement{strings.Join([]string{
 		endorsementHashDelegation,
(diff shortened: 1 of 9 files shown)


experimental/auth-logic/verifier_wrapper_test.go
```diff
@@ -0,0 +1,54 @@
...
+package authlogic
+
+import (
+	"testing"
+)
+
+func TestVerifierWrapper(t *testing.T) {
+	want :=
+		`"OakFunctionsLoader::Verifier" says {
+"OakFunctionsLoader::EndorsementFile" canSay expected_hash("OakFunctionsLoader::ProvenanceFile", any_hash).
```
aferr, line 26: It looks like the endorsement files no longer list a hash for the provenance file. I'll change this shortly so that this statement is no longer here and instead there is another `ProvenanceFile` principal that also gives an expected hash of the binary and I'll change the rule at the bottom to check that the provenance file and the endorsement file give the same hash. (The `ProvenanceFile` principal will be distinct from the `ProvenanceFileBuilder` -- the builder one is the result of trying to build the using the provenance file and the `ProvenanceFile` one just parses a hash from it).

aferr: fixed. @rbehjati ready for review now

experimental/auth-logic/verifier_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+
+type verifierWrapper struct{ appName string }
+
+func (v verifierWrapper) EmitStatement() UnattributedStatement {
+	endorsementPrincipal := fmt.Sprintf("\"%v::EndorsementFile\"", v.appName)
+	provenancePrincipal := fmt.Sprintf("\"%v::Provenance\"", v.appName)
+	binaryPrincipal := fmt.Sprintf("\"%v::Binary\"", v.appName)
+	appPrincipal := fmt.Sprintf("\"%v\"", v.appName)
+
+  // The verifier needs to import expected hashes from both the endorsement
```
rbehjati, line 32: nit: The indentation looks a bit weird!

experimental/auth-logic/verifier_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+  // and provenance files. If we use the same predicate to represent both of 
+  // these statements in this SecPal-based syntax for authorization logic, we 
+  // will lose track of who originated each statement. For example, if we just 
+  // used `Binary has_expected_hash(<hash>)` and the verifier delegates this 
+  // predicate to both the endorsement file and the provenance file, we cannot 
+  // write a policy that looks for the same predicate from both. To work around
+  // this we had a second argument to the predicate to track the original
+  // speaker.
+
+	endorsementHashDelegation :=
```
rbehjati, line 42: Can the strings in this part of the code be defined as consts at the top of the file?

experimental/auth-logic/verifier_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+	provenanceDelegation :=
+		"\"ProvenanceFileBuilder\" canSay any_principal hasProvenance(any_provenance).\n"
+
+	hashMeasurementDelegation :=
+		"\"Sha256Wrapper\" canSay some_object has_measured_hash(some_hash).\n"
+
+	rekorLogCheckDelegation :=
+		"\"RekorLogCheck\" canSay some_object canActAs \"ValidRekorEntry\".\n"
+
+	tab := "    "
```
rbehjati, line 59: Does not `\t` work?

experimental/auth-logic/verifier_wrapper_test.go
```diff
@@ -0,0 +1,53 @@
...
+// Package authlogic contains logic and tests for interfacing with the
+// authorization logic compiler
+package authlogic
+
+import (
+	"testing"
+)
+
+func TestVerifierWrapper(t *testing.T) {
+	want :=
```
rbehjati, line 24: Can this be loaded from a file? It looks a bit chaotic! :)

experimental/auth-logic/verifier_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+type verifierWrapper struct{ appName string }
+
+func (v verifierWrapper) EmitStatement() UnattributedStatement {
+	endorsementPrincipal := fmt.Sprintf("\"%v::EndorsementFile\"", v.appName)
```
tiziano88, line 27: Use raw string literals, and use %s for formatting.

experimental/auth-logic/verifier_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+// Package authlogic contains logic and tests for interfacing with the
+// authorization logic compiler
+package authlogic
+
+import (
+	"fmt"
+	"strings"
+)
+
+type verifierWrapper struct{ appName string }
```
tiziano88, line 24: Please add a comment here.

experimental/auth-logic/verifier_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+
+	hashMeasurementDelegation :=
+		"\"Sha256Wrapper\" canSay some_object has_measured_hash(some_hash).\n"
+
+	rekorLogCheckDelegation :=
+		"\"RekorLogCheck\" canSay some_object canActAs \"ValidRekorEntry\".\n"
+
+	tab := "    "
+
+	binaryIdentificationRule :=
```
tiziano88, line 61: This is getting quite complicated, I suggest looking into https://pkg.go.dev/text/template at some point. In fact, this entire method can probably be a single template, and it should read quite nicely.

experimental/auth-logic/verifier_wrapper.go
aferr: fixed

experimental/auth-logic/verifier_wrapper.go
aferr: I'm just letting `go fmt` decide on that.

experimental/auth-logic/verifier_wrapper.go
aferr: Some of these strings depend on the name in the wrapper, so those can't. Also, this wrapper might move and encapsulating these things in the function is useful in that case.

experimental/auth-logic/verifier_wrapper.go
aferr: I don't really like literal tab characters but I did it anyway.

experimental/auth-logic/verifier_wrapper.go
aferr: Opened [#39]. Left TODO comment.

experimental/auth-logic/verifier_wrapper_test.go
aferr: Fixed.

aferr:
@rbehjati @tiziano88 So far I have been making the requested changes and then hitting "request re-review" even when I have an approval. I figure it's safer to request re-review in case something wasn't quite resolved or wasn't quite what you wanted. Would you actually rather I just not request re-review and merge directly if I think the changes are likely to be what you wanted?

experimental/auth-logic/verifier_wrapper.go
```diff
@@ -0,0 +1,81 @@
...
+		endorsementHashDelegation,
+		provenanceHashDelegation,
+		provenanceDelegation,
+		hashMeasurementDelegation,
+		rekorLogCheckDelegation,
+		binaryIdentificationRule}[:], "\n")}
+}
+
+func (v verifierWrapper) Identify() Principal {
+	return Principal{fmt.Sprintf("\"%v::Verifier\"", v.appName)}
```
tiziano88, line 80: Use raw strings and %s.

tiziano88 (approved):

experimental/auth-logic/verifier_wrapper.go
aferr: Fixed this one. Fine to leave it for the ones that actually do involve `\n` ? I think for those using `"` is slightly shorter. I don't really have a strong preference though.

tiziano88: Yeah I noticed that it would be annoying for the ones with \n so I didn't comment on those. Up to you, maybe the EDSL or the template approach would help with this, no big deal anyways.

aferr:
> 

Ah okay that is good to know. I tend to very eagerly "mark as resolved" under the expectation that the other person will just un-resolve it, but I can leave things unresolved for those cases instead.

rbehjati:
> > 
> 
> Ah okay that is good to know. I tend to very eagerly "mark as resolved" under the expectation that the other person will just un-resolve it, but I can leave things unresolved for those cases instead.

Oh! I had not noticed the "Unresolve conversation" button. Just saw it. Then please keep eagerly marking the conversations as resolved :)

experimental/auth-logic/verifier_wrapper.go
aferr: That string depends on `endorsementPrincipal` which itself depends on `v.appName`, which is state in the `verifierWrapper` struct that the method is implemented on, so that specific string actually can't go outside the function.

aferr: Ah okay. I think that makes it harder to read because then the pieces are more spread out... I can do it that way if you prefer though.

rbehjati: I think these strings have very specific structures, so we don't really need to read the exact string (I could be wrong), so as long as the const has an expressive name the code should be readable.

rbehjati (approved):

experimental/auth-logic/verifier_wrapper.go
aferr: fixed

rbehjati:
Looks like `check-license-lines` failed.

aferr:
Right, I was just looking at that... I must have hit merge before the job was done running by accident... the error does not seem to have to do with the actual licenses https://github.com/project-oak/transparent-release/runs/5741779990?check_suite_focus=true

And it looks like this is already passing again in the next PR that comes right after it:
https://github.com/project-oak/transparent-release/pull/37

So I am not sure what's going on here, but it looks like it's already not a problem.

rbehjati:
Looks like it is because you merged too early, and the PR branch was deleted. I've updated the config, so we can merge only after all the checks are passed :)

PR #43, opened 2022-03-30
Wrapper for Endorsement Files

aferr:
Now that this is done with a  branch from my fork rather than from a branch on the main repository, you will also see the commits from the other PR that this depends on, making it harder to review. (With branches on the main repository, you can set a branches other than main as the base).

To make this somewhat easier for this PR you can look at just the `endorsement_wrapper.go` and `endorsement_wrapper_test` files -- the other things should not have changed since the commits from the other in-flight PR.

### DIFF
--- a/experimental/auth-logic/endorsement_wrapper.go
+++ b/experimental/auth-logic/endorsement_wrapper.go
@@ -59,7 +59,7 @@ type ValidatedEndorsement struct {
 	ExpiryTime  time.Time
 }
 
-// This parses an endorsement file (in JSON) and
+// ParseEndorsementFile parses an endorsement file (in JSON) and
 // produces an `Endorsement` data structure.
 func ParseEndorsementFile(path string) (*Endorsement, error) {
 	endorsementBytes, err := ioutil.ReadFile(path)
--- a/experimental/auth-logic/endorsement_wrapper.go
+++ b/experimental/auth-logic/endorsement_wrapper.go
@@ -17,11 +17,11 @@
 package authlogic
 
 import (
-  "encoding/json"
-  "io/ioutil"
-  "fmt"
-  "regexp"
-  "time"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"time"
 )
 
 type endorsementWrapper struct{ endorsementFilePath string }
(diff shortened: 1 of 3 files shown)


aferr:
I didn't see any data structures to which the endorsement files are parsed into from JSON, so I made some here. If this is what you want to keep, then it probably makes sense to put these data structures in some other part of the repository and refactor. Let me know if this is what you want me to do or if you want something else.

aferr:
It looks like in your `example.json` file example of an endorsement file, you gave the app a different name from the provenance file (the endorsmenet file ends names the app `oak_functions_loader-0f2189703c57845e09d8ab89164a4041c0af0a62` but the provenance file just calls this `oak_functions_loader`). Is this a canonical difference in the naming? I need the generated authorization logic code to use the same names to talk about the same app, so I removed the trailing part in `getShortAppName()`. Let me know if this difference in names is really intentional.

aferr:
This is now probably easier to review since the PR it depended on is merged.

rbehjati:
> It looks like in your `example.json` file example of an endorsement file, you gave the app a different name from the provenance file (the endorsmenet file ends names the app `oak_functions_loader-0f2189703c57845e09d8ab89164a4041c0af0a62` but the provenance file just calls this `oak_functions_loader`). Is this a canonical difference in the naming? I need the generated authorization logic code to use the same names to talk about the same app, so I removed the trailing part in `getShortAppName()`. Let me know if this difference in names is really intentional.

I'd say the subject name in the provenance example should be updated to `oak_functions_loader-0f2189703c57845e09d8ab89164a4041c0af0a62`. We need to be able to link the provenance and the endorsement files using the subject name and digest. Feel free to updated the example, but it is likely the tests need to be updated too.

experimental/auth-logic/endorsement_wrapper.go
```diff
@@ -0,0 +1,153 @@
...
+type Predicate struct {
+	ValidityPeriod ValidityPeriod `json:"validityPeriod"`
+}
+
+type ValidityPeriod struct {
+	ReleaseTime string `json:"releaseTime"`
+	ExpiryTime  string `json:"expiryTime"`
+}
+
+func ParseEndorsementFile(path string) (*Endorsement, error) {
```
rbehjati, line 56: Add some documentation.

experimental/auth-logic/endorsement_wrapper.go
```diff
@@ -0,0 +1,153 @@
...
+	err = json.Unmarshal(endorsementBytes, &endorsement)
+	if err != nil {
+		return nil,
+			fmt.Errorf("could not unmarshal the endorsement file: %v", err)
+	}
+
+	return &endorsement, nil
+}
+
+func (ew endorsementWrapper) getShortAppName() (string, error) {
```
rbehjati, line 73: I think it is better to remove this and update the name in the provenance example to match that in the endorsement example. 
The tooling for generating endorsements from provenances copies the subject from the provenance statement to the endorsement statement, so it is safe to assume that the two files have the same subject.

experimental/auth-logic/endorsement_wrapper.go
```diff
@@ -0,0 +1,153 @@
...
+	match := nameExtractRegex.FindStringSubmatch(appName)
+	if len(match) != 2 {
+		return "", fmt.Errorf("Couldn't extract app name from: %s", appName)
+	}
+
+	return match[1], nil
+}
+
+func (ew endorsementWrapper) EmitStatement() (UnattributedStatement, error) {
+	endorsement, err := ParseEndorsementFile(ew.endorsementFilePath)
```
rbehjati, line 96: This function is a bit long and does two things. It parses the endorsement and validates a few things about it, and then it generates the authorization logic code. I think it would be nice to refactor this function and move each part to a different function. It might help to introduce an intermediate struct for holding a validated endorsement. The first part generates an instance of it, and the second part uses that to generate the authorization logic code.

experimental/auth-logic/endorsement_wrapper_test.go
```diff
@@ -0,0 +1,53 @@
...
+"UnixEpochTime" canSay RealTimeIs(any_time).
+
+}`
+
+	testEndorsementWrapper := endorsementWrapper{
+		endorsementFilePath: testEndorsementPath}
+
+	appName, err := testEndorsementWrapper.getShortAppName()
+	if err != nil {
+		fmt.Errorf("couldn't get short app name: %v", err)
```
rbehjati, line 37:
```suggestion
		t.Fatalf("couldn't get short app name: %v", err)
```
I think.

experimental/auth-logic/endorsement_wrapper_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	speaker := fmt.Sprintf(`"%s::EndorsementFile"`, appName)
+
+	statement, err := EmitStatementAs(Principal{Contents: speaker},
+		testEndorsementWrapper)
+	if err != nil {
+		fmt.Errorf("couldn't get short app name: %v", err)
```
rbehjati, line 44:
```suggestion
		t.Fatalf("couldn't get short app name: %v", err)
```

experimental/auth-logic/endorsement_wrapper.go
rbehjati: Also, looks like currently there is no test that involves both the endorsement and the provenance files. So I think, you can just remove this function. The name in the provenance can be fixed in a separate PR.

rbehjati (approved):

experimental/auth-logic/endorsement_wrapper.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	binaryPrincipal := fmt.Sprintf(`"%s::Binary"`, appName)
+	endorsementWrapperName := fmt.Sprintf(`"%s::EndorsementFile"`, appName)
+
+	hasExpectedHash := fmt.Sprintf(`%s has_expected_hash_from("sha256:%s", %s)`,
+		binaryPrincipal, expectedHash, endorsementWrapperName)
+
+	expirationCondition := fmt.Sprintf(
+		`RealTimeIs(current_time), current_time > %d, current_time < %d`,
```
tiziano88, line 141: It should not matter too much in practice, but usually time intervals are inclusive start and exclusive end.

tiziano88 (approved):

experimental/auth-logic/endorsement_wrapper.go
aferr: You're right that's better. Fixed.

experimental/auth-logic/endorsement_wrapper_test.go
aferr: Yes. Fixed.

experimental/auth-logic/endorsement_wrapper_test.go
aferr: Fixed

experimental/auth-logic/endorsement_wrapper.go
aferr: Fixed.

aferr:
> > I didn't see any data structures to which the endorsement files are parsed into from JSON, so I made some here. If this is what you want to keep, then it probably makes sense to put these data structures in some other part of the repository and refactor. Let me know if this is what you want me to do or if you want something else.
> 
> We have an internal version, but that is not yet synced into this repo. I am planning to do some refactoring in the `slsa` package and this will be part of that.

OK at some point this probably needs to be refactored to work with whatever structs you have.

experimental/auth-logic/endorsement_wrapper.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+type ValidatedEndorsement struct {
+	Name        string
+	Sha256      string
+	ReleaseTime time.Time
+	ExpiryTime  time.Time
+}
+
+// This parses an endorsement file (in JSON) and
```
rbehjati, line 62:
```suggestion
// ParseEndorsementFile parses an endorsement file (in JSON) and
```

aferr:
@rbehjati I saw you just suggested a change. Do you want me to hold off on merging this in case you have more comments?

rbehjati:
> @rbehjati I saw you just suggested a change. Do you want me to hold off on merging this in case you have more comments?

That is the only one for now. I think we have to set up linting as a CI step, and this will be easier :)

aferr:
> > @rbehjati I saw you just suggested a change. Do you want me to hold off on merging this in case you have more comments?
> 
> That is the only one for now. I think we have to set up linting as a CI step, and this will be easier :)

What linter would you use. So far I have been running `go fmt`

rbehjati:
> What linter would you use. So far I have been running `go fmt`

golint ./...

PR #44, opened 2022-04-01
Add CI step for linting

rbehjati:


aferr:
You might need to rebase onto: #45

aferr:
Ah, it looks like you may have already changed those, so I can drop #45 if you'd rather.

rbehjati:
> Ah, it looks like you may have already changed those, so I can drop #45 if you'd rather.

I did not fix everything, for instance renaming `AuthLogicStatement`. So either way one of us has to rebase and fix merge conflicts.

aferr:
> > Ah, it looks like you may have already changed those, so I can drop #45 if you'd rather.
> 
> I did not fix everything, for instance renaming `AuthLogicStatement`. So either way one of us has to rebase and fix merge conflicts.

I ran `golint` on the auth-logic` directory locally, so if you just `git checkout experimental/auth-logic/ -- 362698b31585bba35fbf8a4375f979e410c1990f` (where that hash is head of main), and then rebase onto #45, that subidrectory should lint.

experimental/auth-logic/provenance_build_wrapper_test.go
```diff
@@ -61,4 +63,8 @@ func TestProvenanceBuildWrapper(t *testing.T) {
 		t.Errorf("got:\n%v\nwant:\n%v\n", got, want)
 	}
 
+	// cleanup
+	if err := os.RemoveAll("/tmp/release"); err != nil {
```
aferr, line 67: Oh this is much better :)

aferr (approved):

experimental/auth-logic/provenance_build_wrapper_test.go
rbehjati: Does not seem like it. So I am using `os.TempDir` now. I wonder why it did not fail before on the CI.

tiziano88 (approved):

PR #50, opened 2022-04-07
Generate provenance files

rbehjati:
Updates the `build` command line tool to generate provenance files if the build is successful.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -16,8 +16,10 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"log"
+	"os"
 
 	"github.com/project-oak/transparent-release/build"
 )
@@ -27,9 +29,23 @@ func main() {
 		"Required - Path to a toml file containing the build configs.")
 	gitRootDirPtr := flag.String("git_root_dir", "",
 		"Optional - Root of the Git repository. If not specified, sources are fetched from the repo specified in the config file.")
+	provenancePathPtr := flag.String("provenance_path", "",
+		"Required - Output file name for storing the generated provenance file.")
+
 	flag.Parse()
 
-	if err := build.Build(*buildConfigPathPtr, *gitRootDirPtr); err != nil {
-		log.Fatalf("error when building the binary: %v", err)
+	prov, err := build.Build(*buildConfigPathPtr, *gitRootDirPtr)
+	if err != nil {
+		log.Fatalf("Couldn't build the binary: %v", err)
+	}
+
+	// Write the provenance statement to file.
+	bytes, err := json.Marshal(prov)
+	if err != nil {
+		log.Fatalf("Couldn't marshal the provenance: %v", err)
+	}
+
+	if err := os.WriteFile(*provenancePathPtr, bytes, 0644); err != nil {
+		log.Fatalf("Couldn't write provenance file: %v", err)
 	}
 }
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -25,16 +25,16 @@ import (
 )
 
 func main() {
-	buildConfigPathPtr := flag.String("config", "",
+	buildConfigPath := flag.String("build_config_path", "",
 		"Required - Path to a toml file containing the build configs.")
-	gitRootDirPtr := flag.String("git_root_dir", "",
+	gitRootDir := flag.String("git_root_dir", "",
 		"Optional - Root of the Git repository. If not specified, sources are fetched from the repo specified in the config file.")
-	provenancePathPtr := flag.String("provenance_path", "",
+	provenancePath := flag.String("provenance_path", "",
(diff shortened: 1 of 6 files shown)


common/common.go
```diff
@@ -243,26 +254,74 @@ func (b *BuildConfig) VerifyBinarySha256Hash() error {
...
+		Type:          InTotoStatementV01,
+		Subject:       []slsa.Subject{subject},
+		PredicateType: SLSAPredicateV02,
+		Predicate:     predicate,
+	}, nil
+}
+
+func parseBuilderImageURI(imageURI string) (string, string, error) {
+	// We expect the URI of the builder image to be of the form NAME@DIGEST
+	uRIParts := strings.Split(imageURI, "@")
```
tiziano88, line 314:
```suggestion
	URIParts := strings.Split(imageURI, "@")
```

tiziano88 (approved):

cmd/build/main.go
```diff
@@ -27,9 +29,23 @@ func main() {
 		"Required - Path to a toml file containing the build configs.")
 	gitRootDirPtr := flag.String("git_root_dir", "",
 		"Optional - Root of the Git repository. If not specified, sources are fetched from the repo specified in the config file.")
+	provenancePathPtr := flag.String("provenance_path", "",
```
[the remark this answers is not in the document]
rbehjati: Thanks. Fixed.

common/common.go
rbehjati: Done.

common/common.go
```diff
@@ -243,26 +254,74 @@ func (b *BuildConfig) VerifyBinarySha256Hash() error {
...
 
 	if b.ExpectedBinarySha256Hash != "" && b.ExpectedBinarySha256Hash != binarySha256Hash {
-		return fmt.Errorf("the hash of the output binary does not match the expected binary hash; got %s, want %v",
+		return nil, fmt.Errorf("the hash of the output binary does not match the expected binary hash; got %s, want %v",
 			binarySha256Hash, b.ExpectedBinarySha256Hash)
 	}
 
-	return nil
+	subject := slsa.Subject{
+		// TODO: Get the name as an input in the TOML file.
```
mariaschett, line 274: Do we want to open a issue here?

common/common.go
```diff
@@ -243,26 +254,74 @@ func (b *BuildConfig) VerifyBinarySha256Hash() error {
...
+	return &slsa.Provenance{
+		Type:          InTotoStatementV01,
+		Subject:       []slsa.Subject{subject},
+		PredicateType: SLSAPredicateV02,
+		Predicate:     predicate,
+	}, nil
+}
+
+func parseBuilderImageURI(imageURI string) (string, string, error) {
+	// We expect the URI of the builder image to be of the form NAME@DIGEST
```
mariaschett, line 313: nit: Maybem move this as a toplevel comment to to the function?

common/common_test.go
```diff
@@ -67,6 +67,23 @@ func TestLoadBuildConfigFromProvenance(t *testing.T) {
 	checkBuildConfig(config, t)
 }
 
+func TestParseBuilderImageURI(t *testing.T) {
```
mariaschett, line 70: Nice! :) 

I was thinking if a test failing to parse an imageURI makes sense, too, but I am not too convinced myself.

cmd/build/main.go
```diff
@@ -16,20 +16,36 @@
...
 	"log"
+	"os"
 
 	"github.com/project-oak/transparent-release/build"
 )
 
 func main() {
-	buildConfigPathPtr := flag.String("config", "",
+	buildConfigPath := flag.String("build_config_path", "",
 		"Required - Path to a toml file containing the build configs.")
```
mariaschett, line 29: nit & pre-existing: "a" toml file vs -> "the" toml file --- would we expect there to be more than one toml file in general? Maybe also want to clarify that these are the build configs to build the gitRootDir? Is that correct?

cmd/build/main.go
```diff
@@ -16,20 +16,36 @@
...
-	gitRootDirPtr := flag.String("git_root_dir", "",
+	gitRootDir := flag.String("git_root_dir", "",
 		"Optional - Root of the Git repository. If not specified, sources are fetched from the repo specified in the config file.")
```
mariaschett, line 31: nit: `config file` -> `build config file`? Or not?

mariaschett (approved):
Nice :)

common/common.go
rbehjati: The function is not exposed, so it does not really require documentation. I think this comment is clearer here.

common/common_test.go
rbehjati: Good point. Added a negative test.

rbehjati (commented):
Thanks for the review.

Issue #48, opened 2022-04-06
Can We Take Hyphens Out of Subject Names of Endorsement Files?


aferr:
The current format for endorsement files uses a hyphen in the application name:
https://github.com/project-oak/transparent-release/blob/main/schema/amber-endorsement/v1/example.json

has

`oak_functions_loader-0f2189703c57845e09d8ab89164a4041c0af0a62`

I'm using the app name to define several principal names in the generated auth logic. The authorization logic syntax for principal names and predicate arguments doesn't include `-`, and I don't want to add it because it will probably conflict with numeric minus, which we might want (and the grammar is also fully whitespace agnostic). Can you instead use a different character, like:

`oak_functions_loader_0f2189703c57845e09d8ab89164a4041c0af0a62`

or

`oak_functions_loader:0f2189703c57845e09d8ab89164a4041c0af0a62`

or

`oak_functions_loader.0f2189703c57845e09d8ab89164a4041c0af0a62`

or

`oak_functions_loader0f2189703c57845e09d8ab89164a4041c0af0a62`

aferr:
@rbehjati @tiziano88

aferr:
I meant in general. OK I can remove hyphens from the names before emitting them as auth logic, if we need to support it.

PR #52, opened 2022-04-07
Remove hyphens

aferr:
Fixes https://github.com/project-oak/transparent-release/issues/48

### DIFF
--- a/experimental/auth-logic/top_level.go
+++ b/experimental/auth-logic/top_level.go
@@ -30,7 +30,7 @@ const relationDeclarations = ".decl attribute has_expected_hash_from(hash : Sha2
 // release verification process.
 func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (string, error) {
 
-	endorsementAppName, err := wrappers.GetAppNameFromEndorsement(endorsementFilePath)
+	endorsementAppName, err := SanitizeName(wrappers.GetAppNameFromEndorsement(endorsementFilePath))
 	if err != nil {
 		return "", fmt.Errorf("verifyRelease couldn't get name from endorsement file: %s, error: %v", endorsementFilePath, err)
 	}
@@ -46,7 +46,7 @@ func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (str
 		return "", fmt.Errorf("verifyRelease couldn't get endorsement statement: %v", err)
 	}
 
-	provenanceAppName, err := wrappers.GetAppNameFromProvenance(provenanceFilePath)
+	provenanceAppName, err := SanitizeName(wrappers.GetAppNameFromProvenance(provenanceFilePath))
 	if err != nil {
 		return "", fmt.Errorf("verifyRelease couldn't get app name in provenance file: %v", err)
 	}
--- a/experimental/auth-logic/top_level.go
+++ b/experimental/auth-logic/top_level.go
@@ -30,13 +30,13 @@ const relationDeclarations = ".decl attribute has_expected_hash_from(hash : Sha2
 // release verification process.
 func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (string, error) {
 
-	endorsementAppName, err := SanitizeName(wrappers.GetAppNameFromEndorsement(endorsementFilePath))
+	endorsementAppName, err := wrappers.GetAppNameFromEndorsement(endorsementFilePath)
 	if err != nil {
 		return "", fmt.Errorf("verifyRelease couldn't get name from endorsement file: %s, error: %v", endorsementFilePath, err)
 	}
 	endorsementStatement, err := wrappers.EmitStatementAs(
 		wrappers.Principal{
-			Contents: fmt.Sprintf(`"%s::EndorsementFile"`, endorsementAppName),
+			Contents: fmt.Sprintf(`"%s::EndorsementFile"`, SanitizeName(endorsementAppName)),
 		},
 		wrappers.EndorsementWrapper{
 			EndorsementFilePath: endorsementFilePath,
@@ -46,14 +46,14 @@ func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (str
 		return "", fmt.Errorf("verifyRelease couldn't get endorsement statement: %v", err)
 	}
 
-	provenanceAppName, err := SanitizeName(wrappers.GetAppNameFromProvenance(provenanceFilePath))
+	provenanceAppName, err := wrappers.GetAppNameFromProvenance(provenanceFilePath)
 	if err != nil {
 		return "", fmt.Errorf("verifyRelease couldn't get app name in provenance file: %v", err)
 	}
 
 	provenanceStatement, err := wrappers.EmitStatementAs(
 		wrappers.Principal{
-			Contents: fmt.Sprintf(`"%s::Provenance"`, provenanceAppName),
+			Contents: fmt.Sprintf(`"%s::Provenance"`, SanitizeName(provenanceAppName)),
 		},
 		wrappers.ProvenanceWrapper{FilePath: provenanceFilePath},
 	)
@@ -63,7 +63,7 @@ func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (str
 
 	provenanceBuildStatement, err := wrappers.EmitStatementAs(
 		wrappers.Principal{
-			Contents: fmt.Sprintf(`"%s::ProvenanceBuilder"`, provenanceAppName),
+			Contents: fmt.Sprintf(`"%s::ProvenanceBuilder"`, SanitizeName(provenanceAppName)),
 		},
 		wrappers.ProvenanceBuildWrapper{
 			ProvenanceFilePath: provenanceFilePath,
--- a/experimental/auth-logic/top_level.go
+++ b/experimental/auth-logic/top_level.go
@@ -36,7 +36,7 @@ func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (str
 	}
 	endorsementStatement, err := wrappers.EmitStatementAs(
 		wrappers.Principal{
-			Contents: fmt.Sprintf(`"%s::EndorsementFile"`, SanitizeName(endorsementAppName)),
+			Contents: fmt.Sprintf(`"%s::EndorsementFile"`, wrappers.SanitizeName(endorsementAppName)),
 		},
 		wrappers.EndorsementWrapper{
 			EndorsementFilePath: endorsementFilePath,
@@ -53,7 +53,7 @@ func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (str
 
 	provenanceStatement, err := wrappers.EmitStatementAs(
 		wrappers.Principal{
-			Contents: fmt.Sprintf(`"%s::Provenance"`, SanitizeName(provenanceAppName)),
+			Contents: fmt.Sprintf(`"%s::Provenance"`, wrappers.SanitizeName(provenanceAppName)),
 		},
 		wrappers.ProvenanceWrapper{FilePath: provenanceFilePath},
 	)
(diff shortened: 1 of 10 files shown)


aferr:
This depends on the other in-flight PR I have. You may want to old off on reviewing this until that one is in.

experimental/auth-logic/top_level.go
```diff
@@ -36,7 +36,7 @@ func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (str
 	}
 	endorsementStatement, err := wrappers.EmitStatementAs(
 		wrappers.Principal{
-			Contents: fmt.Sprintf(`"%s::EndorsementFile"`, endorsementAppName),
+			Contents: fmt.Sprintf(`"%s::EndorsementFile"`, wrappers.SanitizeName(endorsementAppName)),
```
tiziano88, line 39: FYI if we used more type safe objects to represent principals and statements (as per https://github.com/project-oak/transparent-release/pull/46#discussion_r845316236), we may only have to do this escaping in a single place (i.e. when converting the principal to string). Currently there is nothing that enforces that names are sanitized correctly across the codebase, it needs to be done for each occurrence.

tiziano88 (approved):

experimental/auth-logic/wrappers/wrappers.go
```diff
@@ -96,3 +97,12 @@ func EmitAuthLogicToFile(authLogic Statement, filepath string) error {
...
 	return writeErr
 }
+
+// SanitizeName takes a value that was parsed from something external to
+// authorizaiton logic and returns a new value that can be used as part of the
+// syntax for principal names and arguments in the authorization logic syntax.
+// At present, it removes hyphens. This is exported because it may be used in
+// the main package to get principal names.
+func SanitizeName(oldName string) string {
+	return strings.ReplaceAll(oldName, "-", "")
```
rbehjati, line 107: In general this may result in name-clashes. Is that something we should worry about, or are these names sufficiently localized?

rbehjati, line 107: For better readability, instead of removing `-`, perhaps replace it with some other permitted character (like `:` or `_`) as you suggested in #48.

rbehjati (approved):

experimental/auth-logic/top_level.go
aferr: I don't think type safety has any effect on whether names are sanitized everywhere or not unless the interface for authorization logic (which is meant to generalize to more than one project) is tied to this one specific project (because what sanitization needs to be done is specific to this project).

Because sanitization is specific to this project, we also need a public constructor for principals that takes a string (`ConstructPrincipal: String -> Principal`) and doesn't sanitize.

Your point is to add `CompoundPrincipal: Principal -> Principal -> Principal`, and your claim is that this ensures sanitization happens everywhere.

Sanitize can either be `Sanitize: String -> String` or it can be `Sanitize: String -> Principal`. Presumably your proposal works by assuming it returns a principal (and that a compound principal is passed to `EmitStatement`).

EmitStatement is `(Principal, Unattributed) -> Statement`. I think your proposal is meant to work by forcing a compound statement to be used in this position, but it doesn't. Even with these functions available, the programmer can still write either:

`EmitStatement(CompoundPrincipal(Sanitize(<AppName>), ConstructPrincipal("::Endorsement")), statement)`

or

`EmitStatement(CompoundPrincipal(ConstructPrincipal(<AppName>), ConstructPrincipal("::Endorsement")), statement)`

or

`EmitStatement(ConstructPrincipal(<AppName> + "::Endorsement"), statement)`

And 2 out of 3 of those don't do sanitization. So your idea does not achieve its stated purpose, and despite not achieving its stated purpose, it does make the code more complicated.

If the goal really is to ensure that sanitization happens everywhere (and in doing so tying the interface to this specific project), a solution that would actually work is to just move the sanitization logic into the implementation of the `String -> Principal` constructor.

experimental/auth-logic/wrappers/wrappers.go
aferr: Is your point that there could be sources with both names like `FooBar` and `Foo-Bar` and this will map both to `FooBar` when they should have been different? If so, you are right, this could cause errors in general. It sounds pretty rare though. I guess a way to fix this is:

Define: s is a sanitizable name (SN) when `s != SanitizeName(s)` and call `SanitizeName(s)` its sanitization, then:
 - maintain a `map`, `sanMap` from sanitizations to the sanitizable names that generated them
 - whenever a sanitizable name `s` is encountered:
     - if `sanMap.get(SanitizeName(s))` is empty, add `sanMap[SanitizeName(s) --> s]`
     - else: check `sanMap.get(s) == s`
         - if equal: do nothing
         - if not equal: there is an erroneous name collision, throw error.
         
Does that solution make sense? Is there a simpler one?

Because this seems rare, I am inclined to open an issue for this rather than fix it now.

experimental/auth-logic/top_level.go
aferr: Hmmm... I guess the sanitization that needs to happen is not all that specific to transparent release. So perhaps it is best to just move the sanitization logic into the constructor of principal and then it really will happen everywhere. WDYT?

experimental/auth-logic/wrappers/wrappers.go
rbehjati: Yes. That is what I mean. I cannot think of a simpler solution right now. Feel free to fix this in a separate PR or even  ignore it, if this situation is very unlikely to happen. But perhaps clarify in the doc comment.

experimental/auth-logic/top_level.go
tiziano88: > Hmmm... I guess the sanitization that needs to happen is not all that specific to transparent release. So perhaps it is best to just move the sanitization logic into the constructor of principal and then it really will happen everywhere. WDYT?

yes that's what I meant; have a type safe constructor for Principal, and then also type constructors for compound principals and statements etc. I think it should be very rare that we pass actual strings around.

> `EmitStatement(CompoundPrincipal(ConstructPrincipal(<AppName>), ConstructPrincipal("::Endorsement")), statement)`
>
> or
>
> `EmitStatement(ConstructPrincipal(<AppName> + "::Endorsement"), statement)`

These should be different principals (if at all); currently, as you noted, the sanitize function is not even injective, so we probably need to somehow escape special characters instead of purely removing them (or we could forbid some entirely, e.g. whitespaces, or `::`, and prevent creating a principal with those in the first place).

rbehjati: I think the idea is to have more custom-defined Go structs so that instead of changing 7-8 files for a behaviour like this, you'd modify only a single file.

experimental/auth-logic/wrappers/provenance_build_wrapper.go
```diff
@@ -70,8 +70,8 @@ func (pbw ProvenanceBuildWrapper) EmitStatement() (UnattributedStatement, error)
 
 	return UnattributedStatement{
 		Contents: fmt.Sprintf("\"%v::Binary\" hasProvenance(\"%v::Provenance\").\n",
-			applicationName, applicationName) +
+			sanitizedAppName, sanitizedAppName) +
```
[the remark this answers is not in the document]
aferr: I thought this was more readable than a single `fmt.Sprintf` because these are separate statements. I did it a different way that is similar to what you suggested for the verification wrapper (using constants) and uses just one sprintf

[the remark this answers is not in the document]
rbehjati: Then I think having each part as a variable with an expressive name, and a top-level `fmt.Sprintf` would be more readable.

experimental/auth-logic/top_level.go
aferr: @tiziano88 nothing about your reply shows how type safety has any affect on whether sanitization takes place. So I am still pretty convinced it does not have this affect.

What you are both describing is more like an E-DSL, which is like a whole separate project that we have talked about in a few other cases. It might have other merits, but I am pretty sure it does not fix the problem you pointed out in this comment :) Do you just want me to make some issue for a go `E-DSL` ?

What about just moving this logic into the constructor for the principal? Nobody commented on that :)

rbehjati: What is E-DSL?

experimental/auth-logic/wrappers/wrappers.go
aferr: fixed

experimental/auth-logic/top_level.go
aferr: Embedded domain specific language. It's a way of implementing something that resembles a programming language by instead implementing a library in some other language, sometimes called the host language. (The E-DSL is "embedded" in the host language). Examples of E-DSLs include tensorflow, [Chisel](https://en.wikipedia.org/wiki/Chisel_(programming_language)), and [Halide](https://en.wikipedia.org/wiki/Chisel_(programming_language)).

experimental/auth-logic/wrappers/wrappers.go
aferr: fixed

experimental/auth-logic/top_level.go
rbehjati: I don't think that is what we are talking about here. My understanding is that we just need some refactoring. 

For instance, you already have a `ValidatedEndorsement` struct with a `Name` field, which is a string. Instead of a string it could be custom-defined `SanitizedName` type that takes a string and stores a sanitized version of it. You'd probably want to have a `ValidatedProvenance` as well, with a `Name` field of the same `SanitizedName` type. Instances of these types are generated when you load the endorsement and the provenance from file, and you use these objects in the rest of the code.

experimental/auth-logic/wrappers/wrappers.go
aferr: https://github.com/project-oak/transparent-release/issues/58

experimental/auth-logic/top_level.go
aferr: @rbehjati Am I correct in understanding that the goal of this proposal is to ensure that all principal names are sanitized? If that is the goal, isn't it better to just call `SanitizeName` inside the constructor of `Principal`? 

If this is just moved inside the constructor, all the names are still sanitized and the code is changed in just one place to do this. By changing the principal constructor to sanitize, the programmer also never has to think about this whereas if the interface of principal is changed to take a new `SanitizedName` type, the programmer needs to still call `SanitizeName()` everywhere.

aferr: I guess since `go` doesn't have constructors, the equivalent thing to do would be to make

```
func ConstructPrincipal(contents String) {
   return Principal{ Contents: sanitizeName(contents) }
}
```

and then in the code, call `ConstructPrincipal` everywhere rather than writing `Principal` directly.

aferr: https://github.com/project-oak/transparent-release/issues/59

rbehjati: > isn't it better to just call SanitizeName inside the constructor of Principal?

Then it is really not `sanitizeName` but `sanitizeString`. Is that what you want? Or is `Contents` exactly the same as `Name`? If so, perhaps it should be renamed to `Name`. 

> the programmer needs to still call SanitizeName() everywhere.

No. The programmer only needs to use `SanitizeName()` in the constructor of the types that have an instance of `SanitizeName`. This is a more type-safe approach in my opinion. With your proposal, there will be an inconsistency between the contents of a `Principal` and a related `ValidatedEndorsement`. It might not be an issue now, but it could become an issue as the code evolves. 

I should add that that I don't have a complete model of all the structs and their relations in mind, so maybe you are right and just sanitizing all strings when creating an instance of `Principal` is good enough. 

Please double-check what is the convention for naming constructors in Go. I think I have seen `new`. Also, `Contents` must be changed to `contents` so that it is not exposed. Otherwise, one could just instantiate `Principal` without calling the constructor.

aferr: If this does not need to be fixed in this PR, and since you both approved the PR, is it OK to move further discussion to that issue and resolve this thread ?

I will speculatively resolve this thread. I added a link to this thread in that issue.

rbehjati: Sounds good to me. Thanks :)

PR #55, opened 2022-04-07
Genrule for Running Auth Logic Verification; Integration Test

aferr:


aferr:
For some reason, building with the provenance file is taking much longer to run than it normally does locally... I'm pushing this now in part to see if it will work here.

aferr:
@rbehjati @tiziano88 Neither of you would happen to have secret insider info about why building is now taking 20m, would you? I think building from provenance files used to take less time than it does now... kind of a long shot, so if you don't have any info that's okay :)

rbehjati:
Yes, I think the build used to take less time. We should perhaps refactor the code so that those large tests do not all actually fetch the Oak repo and build it. That functionality does not need to be checked multiple times. 

But on this PR the CI is not even running the tests. The build step (`"${GITHUB_WORKSPACE}/bin/bazel" build //...`) is failing, and taking 20mins.

aferr:
> Yes, I think the build used to take less time. We should perhaps refactor the code so that those large tests do not all actually fetch the Oak repo and build it. That functionality does not need to be checked multiple times.
> 
> But on this PR the CI is not even running the tests. The build step (`"${GITHUB_WORKSPACE}/bin/bazel" build //...`) is failing, and taking 20mins.

Thanks for looking into this and thinking about it. I think I figured out the problem for this specific PR, and I think I am about to fix it, so you shouldn't need to spend more time on this.

aferr:
@rbehjati Oh but actually, once this is working, there will be two tests in the auth-logic subidirectory that both involve building from provenance files:
 - one is `provenance_build_wrapper_test`
 - the new one in this PR also calls the provenance_build_wrapper

One relatively cheap way to keep the time for CI lower would be to deprecate `provenance_build_wrapper_test` so that just one of these is built. Although deprecating tests is not usually desirable. WDYT ?

experimental/auth-logic/top_level.go
```diff
@@ -87,7 +87,7 @@ func verifyRelease(appName, endorsementFilePath, provenanceFilePath string) (str
 	}
 
 	topLevelQuery := "verification_success = query " + verifierPrincipal.String() +
-		" says \"" + appName + "::Binary\" canActAs " + appName + "?"
+		" says \"" + appName + "::Binary\" canActAs \"" + appName + "\"?"
```
rbehjati, line 90: I think `fmt.Sprintf` with `%q` could solve the quotation problem.

rbehjati (approved):
LGTM.

experimental/auth-logic/oak_auth_logic_verification_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"testing"
+)
+
+func TestSimpleAuthLogic(t *testing.T) {
+	actualQueryValues, err := emitOutputQueries(".")
+	if actualQueryValues == nil || err != nil {
+		t.Fatalf("Could not parse verification query results for oak_functions_loader: %v", err)
+	}
+
+	// In this case verification is expected to fail because the endorsement
```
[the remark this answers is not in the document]
aferr: Oh I meant to actually write `because the endorsement file has *expired*` instead of "failed", so I'll change it to that. Thanks for noticing that this is comment was wrong.

experimental/auth-logic/top_level.go
aferr: fixed

PR #56, opened 2022-04-08
Fix linting 

rbehjati:
Currently the linting does not fail is `golint` fails with error in stderr. This PR updates the linting script to fail in such cases too.


rbehjati:
@aferr I confirmed that moving ` experimental/auth-logic/main.go` and `experimental/auth-logic/top_level.go` to `experimental/auth-logic/cmd/` fixes the `golint` issue, but the `build` fails. `BUILD` and perhaps other things should be updated accordingly. I'll force-merge this now, and leave it to you to fix the structure :)

PR #61, opened 2022-04-11
An integration test that is expected to pass

aferr:
This is a top level integration test that is expected to pass. There's an endorsement file that isn't expired, a provenance file with the same app name as the endorsement file, and some fixes in the verification policy.

### DIFF
--- a/experimental/auth-logic/oak_verification_passing_test.go
+++ b/experimental/auth-logic/oak_verification_passing_test.go
@@ -18,17 +18,14 @@ import (
 	"testing"
 )
 
-func TestOakVerification(t *testing.T) {
+func TestOakVerificationPassing(t *testing.T) {
 	actualQueryValues, err := emitOutputQueries(".")
 	if actualQueryValues == nil || err != nil {
 		t.Fatalf("Could not parse verification query results for oak_functions_loader: %v", err)
 	}
 
-	// In this case verification is expected to fail because the endorsement
-	// file has expired and the names of the app in the endorsement and provenance
-	// files are different.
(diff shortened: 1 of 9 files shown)


experimental/auth-logic/oak_verification_passing_test.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	var expectedQueryValues = map[string]bool{
+		"oak_verification_passing": true,
+	}
+
+	for query, want := range expectedQueryValues {
+		got := actualQueryValues[query]
+		if want != got {
+			t.Fatalf("Query %q failed; want %t got %t.", query, want, got)
```
rbehjati, line 34:
```suggestion
			t.Errorf("Query %q failed; want %t got %t.", query, want, got)
```

rbehjati (approved):

rbehjati:
> @rbehjati Actually, was the example endorsement file deliberately expired for security/safety reasons? Do you want me to use some app name other than oak functions loader?

It is not quite deliberate. The example was added more than month ago, and to show a realistic expiry date it was set to one month after the release date. Oak Functions loader as the app name sounds good to me :)

aferr:
> > @rbehjati Actually, was the example endorsement file deliberately expired for security/safety reasons? Do you want me to use some app name other than oak functions loader?
> 
> It is not quite deliberate. The example was added more than month ago, and to show a realistic expiry date it was set to one month after the release date. Oak Functions loader as the app name sounds good to me :)

Hopefully transparent release will be so successful that this test involving an expiration date 10 years from now will eventually fail.

PR #62, opened 2022-04-12
Fix provenance output path flag

rbehjati:
`flag` does not allow specifying flags as required, so I've added a default value for `provenance_path`. The path is also converted to an absolute path to make sure the file will be generated in the intended location.


cmd/build/main.go
```diff
@@ -30,11 +31,16 @@ func main() {
 		"Required - Path to a toml file containing the build configs.")
 	gitRootDir := flag.String("git_root_dir", "",
 		"Optional - Root of the Git repository. If not specified, sources are fetched from the repo specified in the toml file.")
-	provenancePath := flag.String("provenance_path", "",
-		"Required - Output file name for storing the generated provenance file.")
+	provenancePath := flag.String("provenance_path", "provenance.json",
+		"Optional - Output file name for storing the generated provenance file. Default: provenance.json")
```
mariaschett, line 35: nit: you could define a `defaultFileName := "provenance.json" to keep it in sync in the description + flag.

mariaschett (approved):
LGTM. Paths... :)

cmd/build/main.go
mariaschett: No, no, I meant keeping the string literal `"provenance.json"` in a varibable (e.g., `defaultFileName`), because you use twice (in line 34 +35).

rbehjati: Ah! I see. I feel like it is not worth it, because then I have to import `fmt` and use `fmt.Sprintf` for the help text to reuse the variable. I'll keep it as is for now :)

PR #63, opened 2022-04-12
Add go.mod and go.sum and vetting

rbehjati:
Adds go.mod and go.sum to make the repo look more like a standard go project. This allows using `go vet`, which is added as a CI step. 

The plan is to then experiment with `goreleaser` in the next step.


experimental/auth-logic/wrappers/provenance_wrapper_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"testing"
 )
 
-const schemaExamplePath = "schema/amber-slsa-buildtype/v1/example.json"
+const provenanceExamplePath = "schema/amber-slsa-buildtype/v1/example.json"
```
rbehjati, line 23: There is another const with the same name in another file with the same package. Since it is a redeclaration, `go vet` complains about it. But `bazel test` fails without this declaration, because the other file declaring this name is not included as a dependency in the bazel target for this test. There are a few ways we can solve this issue:
1. Add the other test file as a dependency for this one. 
2. Move the declaration of the const to another file and include that as a dependency in both test files.
3. Merge all test files into a single file. 

I think the last option is the most conventional one. What do you think @aferr?

rbehjati: Most of the Google3 projects that I have seen have both go.mod, and a BUILD file. 

For the internal release process, we use some parts of this repo, so eventually it should be copied to Google3, but perhaps we could remove BUILD files from this repo, and only have them in Google3. The same way we'd do for any third-party repo. WDYT?

aferr: I'm not sure about what's conventional or typical but between the 3 options @rbehjati listed in the first message, 2) makes the most sense to me. It seems useful to split the test files based on what they are about, I think one of your tests that defines this const is just about provenance files whereas the test this thread is commenting on is also related to auth logic. I think going with 3) makes it harder to separate dependencies between these two parts of the project.

Actually just giving this file path a different name like you already did here might be better than options 1-3) anyway. It's really just a test input, so it doesn't need to have the same variable name everywhere it gets used IMO.

rbehjati: The provenance test won't be affected by option 3, because that is in a separate package. The problem is that the const with the same name is defined in two places in the `wrappers` package, once in `provenance_build_wrapper_test.go` and once in `provenance_wrapper_test.go`. My suggestion is to merge these two files, since they are both tests in the `wrappers` package, but leave the other test files untouched.

aferr: Oh ok merging two tests in the wrappers package seems fine. I think I misunderstood the context.

PR #64, opened 2022-04-12
Refactor verify

rbehjati:
Refactor verify to be able to more easily support different types of provenance verification logic.

PR #65, opened 2022-04-13
Some cleanup in auth-logic wrappers

rbehjati:
- Uses `verify` in the `ProvenanceBuildWrapper` (the name should change to `ProvenanceVerifierWrapper`)
- Some cleanup in the BUILD (merges various test targets into one)

See also #64. When we have more types of provenance verifiers then we can use another, possibly light weight verifier in the wrapper.


aferr (approved):

rbehjati:
Thanks for the review. I don't remember having suggested this particular implementation before :)

Issue #69, opened 2022-04-14
Implement additional types of verifiers


rbehjati:
We need to support additional types of verifiers:
1. A metadata verifier, similar to https://github.com/slsa-framework/slsa-verifier
2. Signature verifier

It should be possible to combine verifiers. For instance to build a verifier that both checks metadata and verifies signatures.

PR #70, opened 2022-04-14
Add skeleton for a metadata verifier

rbehjati:
This is PR adds the skeleton for a metadata verifier for Amber provenances (`AmberProvenanceMetadataVerifier`) and calls it in `ProvenanceBuildWrapper`. This does not give a complete implementation for a metadata verifier, and is mainly intended to solve the problem of slow builds.

I have added a few TODO to identify the remaining work. In particular, we have to decide which fields need to be verified in the metadata verifier. The expected values should then be provided as input both to the `verify` command line tool, and to `ProvenanceBuildWrapper`, as well as the choice of the verifier (see #69).

### DIFF
--- a/verify/verify.go
+++ b/verify/verify.go
@@ -94,6 +94,7 @@ type AmberProvenanceMetadataVerifier struct {
 // against the expected values specified in this
 // AmberProvenanceMetadataVerifier instance. Returns an error if any of the
 // values is not as expected. Otherwise returns nil, indicating success.
+// TODO(#69): Check metadata against the expected values.
 func (verifier *AmberProvenanceMetadataVerifier) Verify(provenanceFilePath string) error {
 	provenance, err := slsa.ParseProvenanceFile(provenanceFilePath)
 	if err != nil {
--- a/verify/verify.go
+++ b/verify/verify.go
@@ -83,3 +83,28 @@ func (verifier *ReproducibleProvenanceVerifier) Verify(provenanceFilePath string
 
 	return nil
 }
+
+// AmberProvenanceMetadataVerifier verifies Amber provenances by comparing the
+// content of the provenance predicate against a given set of expected values.
+type AmberProvenanceMetadataVerifier struct {
+	// TODO(#69): Add metadata fields.
+}
+
+// Verify verifies a given Amber provenance file by checking its content
+// against the expected values specified in this
+// AmberProvenanceMetadataVerifier instance. Returns an error if any of the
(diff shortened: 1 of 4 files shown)


aferr:
This looks OK. I am wondering if after this change the `ProvenanceBuildWrapper` is still useful, and perhaps it is best to just deprecate it. It seems like what it checks is similar to what the `ProvenanceWrapper` checks, so we might not need both of these.

Probably, the `ProvenanceBuildWrapper` should eventually be replaced with something that gets the hash from github, at which point it really would have a distinct job from `ProvenanceWrapper`. 

We could just deprecate the current slow `ProvenanceBuildWrapper` until we have the github one, but the change here does solve the CI slowdown issue, so what you have here is also fine.

aferr (approved):

rbehjati:
Perhaps `ProvenanceBuildWrapper` and `ProvenanceWrapper` could be merged into a `ProvenanceVerifierWrapper`. But some verification is still needed, which is not done in `ProvenanceWrapper`.

With this change, it may seem like `ProvenanceBuildWrapper` is not doing any interesting verification. But as I have explained in a TODO (not exactly with this wording), `ProvenanceBuildWrapper` will have an instance of  `ProvenanceVerifier`, and use that for verifying the provenance before emitting a statement. 

To be more accurate, it might be better to also change `hasProvenance` to `hasVerifiedProvenance` in the auth-logic statement.

verify/verify.go
```diff
@@ -83,3 +83,28 @@ func (verifier *ReproducibleProvenanceVerifier) Verify(provenanceFilePath string
...
+	provenance, err := slsa.ParseProvenanceFile(provenanceFilePath)
+	if err != nil {
+		return fmt.Errorf("couldn't load the provenance file from %s: %v", provenanceFilePath, err)
+	}
+
+	if provenance.Predicate.BuildType != common.AmberBuildTypeV1 {
+		return fmt.Errorf("incorrect BuildType: got %s, want %v", provenance.Predicate.BuildType, common.AmberBuildTypeV1)
+	}
+
+	// TODO(#69): Check metadata against the expected values.
```
mariaschett, line 107: Maybe we want to indicate the todo also in the documentation of the func?

mariaschett (approved):
Thank you! :)

verify/verify.go
rbehjati: Thanks. Done.

PR #73, opened 2022-04-29
Towards Rekor Log Wrapper: Functions unpacking log entries and checking signatures

aferr:
This PR is running with the assumption that [this approach for using bazel + go modules](https://github.com/project-oak/transparent-release/pull/72) is fine (or at least fine to allow for forward progress on writing the go code for this wrapper).

Thanks @tiziano88 for pointers about how to unpack some of this, such as (https://github.com/google/ent/blob/b0804661406ee428d2127d7f2d653d6c1b45f766/cmd/ent/cmd/status.go) + help in chat.


experimental/auth-logic/BUILD
```diff
@@ -72,96 +60,122 @@ go_binary(
...
   """,
-  outs = ["simple.auth_logic.dl", "demo_working.csv", "demo_disappointing.csv"]
+    tools = [
+        "@auth-logic-compiler//file",
+        "@souffle-bin//file",
+    ],
+)
+
+go_library(
+    name = "lib",
```
tiziano88, line 164: Is it common to name things `lib`? I have usually seen the main rule of a package being named with the same name as the package itself.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,131 @@
...
+
+	var logEntry models.LogEntry
+
+	err = json.Unmarshal(logEntryBytes, &logEntry)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal the logEntry from file: %s, err: %v", rekorLogFilePath, err)
+	}
+
+	if len(logEntry) != 1 {
+		return nil, fmt.Errorf("For transparent release, Rekor log entries should contain exactly one UUID: %v", logEntry)
```
tiziano88, line 63:
```suggestion
		return nil, fmt.Errorf("for transparent release, Rekor log entries must contain exactly one UUID: %v", logEntry)
```

see https://datatracker.ietf.org/doc/html/rfc2119

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,131 @@
...
+	return rekordEntry, nil
+}
+
+// Verify signature in a rekord entry. In the context where this is used,
+// this will verify the contents of a rekord entry (an endorsement file)
+// against the product team's public key.
+func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) error {
+	publicKey := rekordEntry.RekordObj.Signature.PublicKey.Content
+	// The unused argument is for extra bytes, not an error
+	block, _ := pem.Decode(publicKey)
```
tiziano88, line 108: Does this work (so it's self documenting instead of needing a comment), or does Go complain about unused var?

```suggestion
	block, _bytes := pem.Decode(publicKey)
```

experimental/auth-logic/wrappers/BUILD
```diff
@@ -21,7 +21,7 @@ package(default_visibility = ["//:__subpackages__"])
 go_library(
   name = "wrapper_interface",
   importpath = "github.com/project-oak/transparent-release/experimental/auth-logic/wrappers",
-  srcs = [
+ srcs = [
```
tiziano88, line 24:
```suggestion
  srcs = [
```

tiziano88 (commented):
Thanks, looks neat!

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,131 @@
...
+// And also possibly validating the inclusion proof
+//
+// TODO look into validating the inclusion proof with
+// logEntryAnon.Verification.ValidateInclusionProof.
+// I'm not sure what to pass for "formats" yet
+// https://github.com/sigstore/rekor/blob/e0b79164f2279ad7c3e723a3ee16afbfcf271188/pkg/generated/models/log_entry.go#L365
+
+func getLogEntryAnonFromFile(rekorLogFilePath string) (*models.LogEntryAnon, error) {
+	// get LogEntry, which is a map from strings to LogEntryAnons
+	logEntryBytes, err := ioutil.ReadFile(rekorLogFilePath)
```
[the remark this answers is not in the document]
aferr: fixed

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not read the rekor log file: %v", err)
+	}
+
+	var logEntry models.LogEntry
+
+	err = json.Unmarshal(logEntryBytes, &logEntry)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal the logEntry from file: %s, err: %v", rekorLogFilePath, err)
```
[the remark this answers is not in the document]
aferr: no longer relevant

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: fixed

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,131 @@
...
+	if !ok {
+		return nil, fmt.Errorf("could not convert NewEntry into rekord. NewEntry: %v,", entryImpl)
+	}
+	return rekordEntry, nil
+}
+
+// Verify signature in a rekord entry. In the context where this is used,
+// this will verify the contents of a rekord entry (an endorsement file)
+// against the product team's public key.
+func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) error {
```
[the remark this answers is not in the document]
aferr: fixed

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,131 @@
...
+	entryImpl, err := types.NewEntry(proposedEntry)
+	if err != nil {
+		return nil, fmt.Errorf("could not convert ProposedEntry into NewEntry: %v, %s", proposedEntry, err)
+	}
+	// rekord
+	rekordEntry, ok := entryImpl.(*rekord.V001Entry)
+	if !ok {
+		return nil, fmt.Errorf("could not convert NewEntry into rekord. NewEntry: %v,", entryImpl)
+	}
+	return rekordEntry, nil
```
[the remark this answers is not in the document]
aferr: Maybe we can do this when we know we need it as described in this newly created issue: #74 ?

I'm confused why we would expect other types. Since these are rekor logs of endorsement files which are specific to our software, can't we just choose which one format we use in all situations? Why do we need this flexibility?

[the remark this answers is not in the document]
aferr: If we eventually want to use `hashedRekord` instead of `rekord` we could just change the one format we use in this code.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -101,31 +104,32 @@ func getRekordEntryFromAnon(logEntryAnon models.LogEntryAnon) (*rekord.V001Entry
...
 	ecdsaKey, ok := pubKeyDecoded.(*ecdsa.PublicKey)
 	if !ok {
-		return fmt.Errorf("public key is not ecdsa: %v", pubKeyDecoded)
+		return nil, fmt.Errorf("public key is not ecdsa: %v", pubKeyDecoded)
 	}
 
 	data, err := hex.DecodeString(*rekordEntry.RekordObj.Data.Hash.Value)
 	if err != nil {
-		return fmt.Errorf("could not decode hash of data: %v", rekordEntry.RekordObj.Data.Hash.Value)
+		return ecdsaKey, fmt.Errorf("could not decode hash of data: %v", rekordEntry.RekordObj.Data.Hash.Value)
```
tiziano88, line 124: Please don't return the public key unless it was actually verified (also clarify this in the function comment); this is guaranteed to be misused, since there is no valid use for a key that did not actually sign the payload that it claims to have signed.

experimental/auth-logic/BUILD
aferr: This rule was autogenerated by gazelle. It's gone now.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if len(logEntry) != 1 {
+		return nil, fmt.Errorf("for transparent release, Rekor log entries must contain exactly one UUID: %v", logEntry)
+	}
+
+	var logEntryAnon models.LogEntryAnon
+	// set logEntryAnon to the only value in LogEntry (which is a map)
+	for _, anon := range logEntry {
+		logEntryAnon = anon
+	}
```
tiziano88, line 73: I would add a `break` here to make it super clear that this is not actually iterating apart from the first entry

aferr:
@tiziano88 since this PR makes a large change to the WORKSPACE and go module files, should I wait until @rbehjati is back and is OK with this?

I can also make forward progress by writing PRs that depend on this one, so that would not be blocking.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Verify signature in a rekord entry. In the context where this is used,
+// this will verify the contents of a rekord entry (an endorsement file)
+// against the product team's public key. It returns the public key if one can
+// be parsed from the entry.
+func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey, error) {
+	publicKey := rekordEntry.RekordObj.Signature.PublicKey.Content
+	// The unused argument is for extra bytes, not an error
+	block, _ := pem.Decode(publicKey)
```
tiziano88, line 112: (I'm well aware this is more of a critique of my own code) The naming seems inconsistent, I suggest renaming variables to be more consistent, e.g. publicKey, publicKeyBlock, publicKeyDecoded, ecdsaPublicKey, etc.

experimental/auth-logic/wrappers/rekor_wrapper_test.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		t.Fatalf("couldn't parse rekor log entry from path: %s. %v", testRekorLogPath, err)
+	}
+
+	rekordEntry, err := getRekordEntryFromAnon(*logEntryAnon)
+	if err != nil {
+		t.Fatalf("couldn't get rekordEntry from body of logEntryAnon. rekordEntry: %v, logEntryAnon: %v, rekordLogFilePath: %s. err: %v", *rekordEntry, *logEntryAnon, testRekorLogPath, err)
+	}
+
+	_, err = verifyRekordLogSignature(rekordEntry)
```
tiziano88, line 34: Could you also validate the public key, since that's the main output?

tiziano88 (approved):

tiziano88:
> @tiziano88 since this PR makes a large change to the WORKSPACE and go module files, should I wait until @rbehjati is back and is OK with this?
> 
> I can also make forward progress by writing PRs that depend on this one, so that would not be blocking.

I think it's fine to go ahead, since either way we need either the current WORKSPACE file listing all the deps, or anyways we may just get rid of it separately.

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: But it's not "verifying the public key", it's verifying the signature against the public key...  but ok will do.

experimental/auth-logic/wrappers/rekor_wrapper_test.go
aferr: It's not the main output. This is just code meant to write the auth logic wrapper. The auth logic wrapper does not need this public key for anything.

tiziano88: wait, how is it possible that the public key is not needed for anything? am I missing something?

experimental/auth-logic/wrappers/rekor_wrapper.go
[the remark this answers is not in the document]
aferr: OK I split.

experimental/auth-logic/wrappers/rekor_wrapper.go

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: fixed.

experimental/auth-logic/wrappers/rekor_wrapper_test.go
aferr: The auth logic wrapper just needs to say that it validated the signature. So it does need the public key just internally to the function `verifyRekordLogSignature`, which only needs to return a bool saying whether the (key, data, signature) triple was valid (basically the result of `ecdsa.VerifyASN1`). Since we are not using the `principals=keys` feature of auth logic, there is no use for this key other than to check that signature.

aferr: Unless the idea is that we also apply the product team key as an input (to this whole `auth-logic-transparent-release` program) and check that input against the one fetched here?

tiziano88: Yeah sorry my wording was poor, I did mean validate the signature, and return the public key (you are right that if the key was invalid in the first place, the signature would not check out -- one hopes)

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -0,0 +1,139 @@
...
+func getLogEntryAnonFromFile(rekorLogFilePath string) (*models.LogEntryAnon, error) {
+	// get LogEntry, which is a map from strings to LogEntryAnons
+	logEntryBytes, err := ioutil.ReadFile(rekorLogFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("could not read the rekor log file: %v", err)
+	}
+	return getLogEntryAnonFromBytes(logEntryBytes)
+}
+
+func getLogEntryAnonFromBytes(logEntryBytes []byte) (*models.LogEntryAnon, error) {
```
rbehjati, line 57: Does not rekor/sigstore already provide this functionality?

go.mod
```diff
@@ -9,6 +9,53 @@ require (
 )
 
 require (
+	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d // indirect
```
rbehjati, line 12: I am surprised that we have to specify all these indirect dependencies. Why are these needed?

experimental/auth-logic/wrappers/rekor_wrapper_test.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		t.Fatalf("couldn't get entryImpl from body of logEntryAnon logEntryAnon: %v, rekordLogFilePath: %s. err: %v", *logEntryAnon, testRekorLogPath, err)
+	}
+
+	rekordEntry, err := getRekordEntryFromEntryImpl(*entryImpl)
+	if err != nil {
+		t.Fatalf("couldn't get rekordEntry from entryImpl. entryImpl: %v, rekordLogFilePath: %s. err: %v", *entryImpl, testRekorLogPath, err)
+	}
+
+	_, err = verifyRekordLogSignature(rekordEntry)
```
rbehjati, line 39: I suppose this will be repeated, more or less as-is, in the implementation of the wrapper. I'd suggest wrapping all this in a higher-level function (named for instance `verifyRekorLogEntry`) and writing the test for that.

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: Not that I could find.

go.mod
aferr: This was just the result of doing `go get github.com/sigstore/rekor`. Do you have another way to get the sigstore library working?

experimental/auth-logic/wrappers/rekor_wrapper_test.go
rbehjati: The test scenario in #77 seems to have some redundancy, but I only had a quick look at the code. Ideally, we want a test for `EmitStatement`, and perhaps additional, detailed tests for some of the functions called in `EmitStatement`. 

Also, the code in `EmitStatement` is repeating the code in the test here. This is the part that I think can be refactored into a single function, tested here, and called in `EmitStatement` in #77.

go.mod
rbehjati: I think I only needed to run `go mod tidy` to get the required dependencies added to the `go.mod`.

experimental/auth-logic/wrappers/rekor_wrapper_test.go
aferr: Since this PR is already merged, what do you want me to do?

In #77 the wrapper also makes the authorization logic statements next to the go code so that it's easier to follow what the policy statement means. If we pack this into a single `verifyRekorLogEntry` we can't do that anymore.

rbehjati: > If we pack this into a single verifyRekorLogEntry we can't do that anymore.

I think it should still be possible, but let me have a closer look at #77 (possibly tomorrow). I think this change can be done as part of #77, or perhaps a separate PR after that.

PR #77, opened 2022-05-03
Rekor log wrapper

aferr:


experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+		return UnattributedStatement{}, fmt.Errorf("couldn't validate logEntryAnon (which includes inclusion proof checking):%v ", err)
+	}
+	inclusionProofStatement := fmt.Sprintf("hasValidInclusionProof(%v).", logEntryPrincipal)
+
+	// Check that the product team public key in the log entry matches
+	// the input public key
+	err = checkEntryPubKeyMatchesExpectedKey(rekordEntry, rlw.productTeamKeyBytes)
+	if err != nil {
+		return UnattributedStatement{}, fmt.Errorf("rekord entry key does not match input product team key: %v, %v, %v", rekordEntry, rlw.productTeamKeyBytes, err)
+	}
```
rbehjati, line 237: I think most of this chunk of code can be wrapped into a separate function (e.g., `verifyRekorLogEntry`), called here and tested in `rekor_wrapper_test.go`. I believe the lines that generate auth logic statements could all be moved to after the call to `verifyRekorLogEntry`. So this function first verifies everything, and only generates all the statements if the verification passes.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+	entryImpl, err := getEntryImplFromAnon(*logEntryAnon)
+	if err != nil {
+		return UnattributedStatement{}, fmt.Errorf("couldn't get entryImpl from body of logEntryAnon %v: %v", *logEntryAnon, err)
 	}
 
-	return ecdsaKey, nil
+	rekordEntry, err := getRekordEntryFromEntryImpl(*entryImpl)
+	if err != nil {
+		return UnattributedStatement{}, fmt.Errorf("couldn't get rekordEntry from entryImpl:, %v %v", *entryImpl, err)
```
rbehjati, line 212:
```suggestion
		return UnattributedStatement{}, fmt.Errorf("couldn't get rekordEntry from entryImpl: %v, %v", *entryImpl, err)
```
```suggestion
		return UnattributedStatement{}, fmt.Errorf("couldn't get rekordEntry from entryImpl:, %v %v", *entryImpl, err)
```

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
 
-	ok = ecdsa.VerifyASN1(ecdsaKey, data, sig)
-	if !ok {
-		return nil, fmt.Errorf("could not verify ecdsa signature. key:%v, data:%v, sig:%v ", ecdsaKey, data, sig)
+	// Unpack rekord log entry from bytes into go structs
+	logEntryAnon, err := getLogEntryAnonFromBytes(rlw.rekorLogEntryBytes)
+	if err != nil {
+		return UnattributedStatement{}, fmt.Errorf("couldn't parse rekor log entry from bytes: %v, %v", rlw.rekorLogEntryBytes, err)
+	}
+	entryImpl, err := getEntryImplFromAnon(*logEntryAnon)
```
rbehjati, line 205: I think most of the code from this line to line 237 can be wrapped into a separate function (e.g., `verifyRekorLogEntry`), called here and tested in `rekor_wrapper_test.go`. I believe the lines that generate auth logic statements could all be moved to after the call to `verifyRekorLogEntry`. So `EmitStatement` first verifies everything, and only generates all the statements if the verification passes.

rbehjati, line 205: I think most of the code from this line to line 237 can be wrapped into a separate function (e.g., `verifyRekorLogEntry`), called here and tested in `rekor_wrapper_test.go`. I believe the lines that generate auth logic statements could all be moved to after the call to `verifyRekorLogEntry`. So `EmitStatement` first verifies everything, and only generates all the statements if the verification passes.

experimental/auth-logic/wrappers/rekor_wrapper.go
rbehjati, line 212:
```suggestion
		return UnattributedStatement{}, fmt.Errorf("couldn't get rekordEntry from entryImpl %v: %v", *entryImpl, err)
```

rbehjati, line 212:
```suggestion
		return UnattributedStatement{}, fmt.Errorf("couldn't get rekordEntry from entryImpl %v: %v", *entryImpl, err)
```
```suggestion
		return UnattributedStatement{}, fmt.Errorf("couldn't get rekordEntry from entryImpl:, %v %v", *entryImpl, err)
```

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+		return UnattributedStatement{}, fmt.Errorf("couldn't get entryImpl from body of logEntryAnon %v: %v", *logEntryAnon, err)
 	}
 
-	return ecdsaKey, nil
+	rekordEntry, err := getRekordEntryFromEntryImpl(*entryImpl)
+	if err != nil {
+		return UnattributedStatement{}, fmt.Errorf("couldn't get rekordEntry from entryImpl:, %v %v", *entryImpl, err)
+	}
+
+	// Authorization logic statmenets are defined next to the go code that
```
rbehjati, line 215: I don't see much readability benefit in this way of organising the code :)

rbehjati, line 215: I don't see much readability benefit in this way of organising the code :)

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+	}
+	inclusionProofStatement := fmt.Sprintf("hasValidInclusionProof(%v).", logEntryPrincipal)
+
+	// Check that the product team public key in the log entry matches
+	// the input public key
+	err = checkEntryPubKeyMatchesExpectedKey(rekordEntry, rlw.productTeamKeyBytes)
+	if err != nil {
+		return UnattributedStatement{}, fmt.Errorf("rekord entry key does not match input product team key: %v, %v, %v", rekordEntry, rlw.productTeamKeyBytes, err)
+	}
+	pubKeyMatchStatement := fmt.Sprintf("hasCorrectPubkey(%v).", logEntryPrincipal)
```
rbehjati, line 238: I guess what we care most about is the identity of the signer. Is not it?

rbehjati, line 238: I guess what we care most about is the identity of the signer. Is not it?

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+
+	//TODO(#76): check that hash of endorsement file matches hash
+	//of rekor log entry contents. To be able to test this, we need a new
```
rbehjati, line 241: Could not we use the same example in https://github.com/project-oak/oak/tree/main/oak_functions/client/rust/testdata?

rbehjati, line 241: Could not we use the same example in https://github.com/project-oak/oak/tree/main/oak_functions/client/rust/testdata? I can share the public key if that is what is missing.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+	//rekor log entry that matches the endorsement file.
+	contentsMatchStatement := fmt.Sprintf("contentsMatch(%v, %v).", logEntryPrincipal, endorsementPrincipal)
+
+	// This is the policy for claiming an endorsement is a valid rekor log entry
+	// which just collects the evidence above into the more compact statement
+	// that the verifier wrapper uses.
+	rekorEntryPolicy := fmt.Sprintf("%v canActAs ValidRekorEntry :- hasValidBodySignature(%v), hasValidInclusionProof(%v), hasCorrectPubKey(%v), contentsMatch(%v, %v).", endorsementPrincipal, logEntryPrincipal, logEntryPrincipal, logEntryPrincipal, logEntryPrincipal, endorsementPrincipal)
```
rbehjati, line 248: nit: For readability, I'd suggest breaking it into multiple lines, with each argument in a separate line.

rbehjati, line 248: nit: For readability, I'd suggest breaking it into multiple lines, with each argument in a separate line.

experimental/auth-logic/wrappers/rekor_wrapper_test.go
```diff
@@ -15,29 +15,86 @@
...
+
+	"github.com/go-openapi/strfmt"
 )
 
 const testRekorLogPath = "experimental/auth-logic/test_data/rekor_entry.json"
+const testPubKeyPath = "experimental/auth-logic/test_data/product_team_key.pub"
+const testUnexpiredEndorsementFilePath = "experimental/auth-logic/test_data/oak_endorsement.json"
 
 func TestRekoLogWrapper(t *testing.T) {
 	logEntryAnon, err := getLogEntryAnonFromFile(testRekorLogPath)
```
rbehjati, line 29: All of these functions are already called in `EmitStatement` via `EmitStatementAs`, so they seem redundant to me. If the idea is to have a unit test for each function, then each test should be a separate function: e.g., `func TestGetLogEntryAnonFromFile((t *testing.T) {...}`.

rbehjati, line 29: All of these functions are already called in `EmitStatement` via `EmitStatementAs`, so they seem redundant to me. If the idea is to have unit-tests for each function, then each should be a separate function, e.g., `func TestGetLogEntryAnonFromFile((t *testing.T) {`.

rbehjati (commented):
Sorry for the late review.

experimental/auth-logic/test_data/product_team_key.pub
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
```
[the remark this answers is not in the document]
aferr: In #79 (which depends on 77) `product_team_key.pub` gets replaced with with `oak_ec_public.pem`. Is it OK to leave it as-is in this PR since it gets changed in the later one (and to avoid merge conflicts)?

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: The example you linked doesn't work because the endorsement file in that example has expired, so checking it in this process would cause an error. The rekord entry was also generated using the expired endorsement file.

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: OK

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: I think so too... Is this comment meant to request a change to the code? I am not sure what change that is if so :)

This auth logic statement is meant to correspond to the check that the pubkey in the rekord entry (which we also use for verifying the signature in the entry) is the same as another input pubkey for the product team that we trust. This check is the same as checking for "the correct identity of the signer". Was your comment a request to phrase this differently from "hasCorrectPubkey" ?

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: duplicate comment.

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: hmmm... in the past I was also requested to NOT break statements into multiple lines (even though I actually would prefer to do what you are suggesting :). However, it looks like the style guide really does say NOT to break expressions into multiple lines [here](go/go-style/guide#line-length).

experimental/auth-logic/wrappers/rekor_wrapper_test.go
aferr: removed these.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -111,9 +120,30 @@ func getRekordEntryFromEntryImpl(entryImpl types.EntryImpl) (*rekord.V001Entry,
 // against the product team's public key. It returns the public key if and only
 // if the signature is valid
 func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey, error) {
-	publicKey := rekordEntry.RekordObj.Signature.PublicKey.Content
+	publicKeyBytes := rekordEntry.RekordObj.Signature.PublicKey.Content
+	ecdsaKey, err := pubKeyBytesToECDSA(publicKeyBytes)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse ecdsa key from rekord entry")
```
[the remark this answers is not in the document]
rbehjati: I've noticed that, but do we want to use this terminology in the error message? `LogEntry in Rekor` might be better, but I would not insist. It really depends on how familiar the audience of this error message are with the implementation details of the data structures in Rekor.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+		return fmt.Errorf("logEntryAnon did not have inclusion proof: %v", logEntryAnon)
+	}
+	return logEntryAnon.Validate(strfmt.Default)
+}
+
+// checkEntryPubKeyMatchesExpectedKey compares the public key of the product
+// team in the Rekor log entry to the key of the product team passed as an
+// input to this wrapper. It returns an error if they are not equal
+// (or if valid keys could not be constructed)
+func checkEntryPubKeyMatchesExpectedKey(rekordEntry *rekord.V001Entry, prodTeamKeyBytes []byte) error {
```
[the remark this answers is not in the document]
rbehjati: Yes, but it does not answer my question! 

I am not sure the public key is included in the entry just to make it self-contained. If that is the case, we could just  ignore that and directly use the product team's key to verify the signature. Would not that be clearer? 

I actually think the public key is included in the log entry to allow verifying the identity of the signer, in a general manner. So, I think it is better to explicitly reflect this in the code.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -122,18 +152,107 @@ func verifyRekordLogSignature(rekordEntry *rekord.V001Entry) (*ecdsa.PublicKey,
...
+	logECDSAPubKey, err := pubKeyBytesToECDSA(rekordEntry.RekordObj.Signature.PublicKey.Content)
```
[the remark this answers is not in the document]
rbehjati: Makes sense. Thanks. Can you please add a clarifying comment to the code, if it is not already there?

experimental/auth-logic/wrappers/rekor_wrapper.go
rbehjati: My comment is aimed at clarifying the purpose of this statement in the auth logic code :)
 
I'd suggest rephrasing it to be clear that here we are checking the identity of the signer, if it makes sense from the auth logic policy perspective.

experimental/auth-logic/wrappers/rekor_wrapper.go
rbehjati: The style guide you've linked gives a 404. I'd be surprised to see that as the advice. What I remember from an earlier discussion is that `Go` is not picky about the length of the line, but the advice is to break a line, if it is too long. And this line is really too long; particularly seeing that the return statement after it is nicely broken into short lines, with each argument on one line.

rbehjati (commented):
I am pretty sure I did not leave duplicate comments. I blame GitHub for this.

experimental/auth-logic/wrappers/rekor_wrapper.go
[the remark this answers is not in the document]
aferr: If we check that the key in the rekor entry is the same as an expected product team key, it doesn't matter which one we use to check the signature (since they are the same key!)

We also should check that the public key in the rekor entry is the same as the product team key because if it isn't then the signature cant be verified with the key in the entry (which is suspicious because that's not how it should work), and it also suggests the party making the entry is not the one with the trusted product team key.

experimental/auth-logic/wrappers/rekor_wrapper.go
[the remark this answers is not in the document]
aferr: Good point about the error message. I also changed the name of the publicly exported function.

experimental/auth-logic/wrappers/rekor_wrapper.go

experimental/auth-logic/wrappers/rekor_wrapper.go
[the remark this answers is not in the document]
rbehjati: I am not entirely convinced. I think we have to figure out why exactly Rekor includes the public key in the LogEntry, and what use cases they have in mind. Because, exactly as you described above, just validating the signature against the public key in the LogEntry is not enough. 

But we can come back to this in a separate PR.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -186,55 +180,71 @@ func checkEntryPubKeyMatchesExpectedKey(rekordEntry *rekord.V001Entry, prodTeamK
...
+
+	// The generated authorization logic statements correspond to what `VerifyRekordEntry` checks.
+
+	// Verify rekor log entry signature
+	logEntrySignatureStatement := fmt.Sprintf("hasValidBodySignature(%v).", logEntryPrincipal)
+
+	// Verify inclusion proof
+	inclusionProofStatement := fmt.Sprintf("hasValidInclusionProof(%v).", logEntryPrincipal)
+
+	// Check that the product team public key in the log entry matches
```
rbehjati, line 254: supernit: The line-break in the comment is not needed I think.

rbehjati (approved):
Thanks :)

Issue #76, opened 2022-05-03
Check hash in rekor log entries matches hash of endorsement file in wrapper


aferr:
To be able to test this, we need a new Rekor log entry that matches the unexpired endorsement file used in tests.

PR #78, opened 2022-05-04
Check hash of endorsement against rekor entry

aferr:
closes https://github.com/project-oak/transparent-release/issues/76


experimental/auth-logic/test_data/product_team_key.pub
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
```
rbehjati, line 1: What is the difference between this key and the oak_ec_public.pem key? Is not the product team Oak?

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -186,6 +187,14 @@ func checkEntryPubKeyMatchesExpectedKey(rekordEntry *rekord.V001Entry, prodTeamK
 	return nil
 }
 
+func compareEndorsementAndRekordHash(rekordEntry *rekord.V001Entry, endorsementBytes []byte) error {
+	endorsementHash := fmt.Sprintf("%x", sha256.Sum256(endorsementBytes))
+	if endorsementHash != *rekordEntry.RekordObj.Data.Hash.Value {
+		return fmt.Errorf("Hash values of endorsement bytes and rekor dentry not equal. endorsementBytes: %v, endorsementHash: %s, rekordEntry: %v, rekordHash: %v", endorsementBytes, endorsementHash, rekordEntry, *rekordEntry.RekordObj.Data.Hash.Value)
```
rbehjati, line 193: The error message looks very verbose. Perhaps just print the expected hash and the actual one with the `got/want` format.

experimental/auth-logic/test_data/product_team_key.pub
aferr: I don't have the Oak private key you used in the past. Since this is just a test, it doesn't matter if we use the actual keys you expect to use in real life. (In fact, it may be safer this way :)

I needed to have *some* private key to test with because:
 - I needed to make a non-expired endorsement file. the one you've been using expired back in march
 - I needed to make a rekor entry that uses the non-expired endorsement file
 - I needed to use some private key to create the rekor entry.

Does that make sense?

experimental/auth-logic/test_data/oak_ec_private.pem
```diff
@@ -0,0 +1,8 @@
+-----BEGIN EC PARAMETERS-----
```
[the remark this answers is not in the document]
tiziano88: Well this key is not actually secret, so it is fine (in fact expected) to be checked in. Otherwise how would we use it to sign additional things for testing?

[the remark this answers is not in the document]
tiziano88: If we want to try something different in the json files, we would have to sign them again according to the instructions in experimental/auth-logic/test_data/rekor_entry_instructions.md , and we wouldn't want to regenerate the keys just for that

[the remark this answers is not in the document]
rbehjati: Sounds good.

experimental/auth-logic/test_data/product_team_key.pub
aferr: Ah sorry, I misunderstood. I thought I renamed the key but I guess I did not delete the old one. It's fixed now. Thanks.

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: OK I made it just include the two hashes. I thought `got/want` should be just for tests, so I left it saying `endorsement hash: %v, rekor hash: %v`, since if I were reading the message I wouldn't be sure which one is got and which one is want.

If you still would rather it be got/want, which should be "got"? The endorsement one?

rbehjati: Makes sense. Thanks. I don't particularly like the `got/want` style, but I think I've seen it used in code other than the tests (I could be wrong). Your wording is clearer in my opinion :)

experimental/auth-logic/test_data/rekor_entry_instructions.md
```diff
@@ -0,0 +1,16 @@
+The `rekor_entry.json` file was generated by doing the following
```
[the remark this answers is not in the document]
aferr: #80

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -185,10 +186,22 @@ func checkEntryPubKeyMatchesExpectedKey(rekordEntry *rekord.V001Entry, prodTeamK
 	return nil
 }
 
+// compareEndorsementAndRekordHash compares the hash referenced in a
+// rekord entry to the hash of input bytes.  For this use-case, the input
+// bytes will be an endorsement file
+func compareEndorsementAndRekordHash(rekordEntry *rekord.V001Entry, endorsementBytes []byte) error {
```
rbehjati, line 192:
```suggestion
func compareEndorsementAndRekorHash(rekordEntry *rekord.V001Entry, endorsementBytes []byte) error {
```

nit: I think we decided to user rekor.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -185,10 +186,22 @@ func checkEntryPubKeyMatchesExpectedKey(rekordEntry *rekord.V001Entry, prodTeamK
 	return nil
 }
 
+// compareEndorsementAndRekordHash compares the hash referenced in a
+// rekord entry to the hash of input bytes.  For this use-case, the input
+// bytes will be an endorsement file
```
rbehjati, line 191:
```suggestion
// bytes will be an endorsement file.
```

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -223,6 +236,17 @@ func VerifyRekorEntry(rekorLogEntryBytes []byte, productTeamKeyBytes []byte) err
 		return fmt.Errorf("rekord entry key does not match input product team key: %v, %v, %v", rekordEntry, productTeamKeyBytes, err)
 	}
 
+	// Check that hash of endorsement file matches hash of rekor log entry
```
rbehjati, line 239: nit: no line break?

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -185,10 +186,22 @@ func checkEntryPubKeyMatchesExpectedKey(rekordEntry *rekord.V001Entry, prodTeamK
...
+			endorsementHash, *rekordEntry.RekordObj.Data.Hash.Value)
+	}
+	return nil
+}
+
 // VerifyRekorEntry verifies a rekord entry by checking that the signature
 // it includes is valid, that the inclusion proof is valid, and that it
 // was created using a public key for the product team that we trust.
-func VerifyRekorEntry(rekorLogEntryBytes []byte, productTeamKeyBytes []byte) error {
+func VerifyRekorEntry(rekorLogEntryBytes []byte, productTeamKeyBytes []byte, endorsementFilePath string) error {
```
rbehjati, line 204: I'd suggest passing the endorsement bytes to this function to be consistent with rest of the arguments.

experimental/auth-logic/wrappers/rekor_wrapper.go
```diff
@@ -238,12 +262,12 @@ func (rlw RekorLogWrapper) EmitStatement() (UnattributedStatement, error) {
 	endorsementPrincipal := fmt.Sprintf(`"%s::EndorsementFile"`, SanitizeName(endorsementAppName))
 	logEntryPrincipal := fmt.Sprintf(`"%s::RekorLogEntry"`, SanitizeName(endorsementAppName))
 
-	err = VerifyRekorEntry(rlw.rekorLogEntryBytes, rlw.productTeamKeyBytes)
+	err = VerifyRekorEntry(rlw.rekorLogEntryBytes, rlw.productTeamKeyBytes, rlw.endorsementFilePath)
```
rbehjati, line 265: I'd suggest including the endorsement bytes in RekorLogWrapper, to be more consistent throughout.

rbehjati (approved):

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: fixed

experimental/auth-logic/wrappers/rekor_wrapper.go
aferr: Fixed

experimental/auth-logic/wrappers/rekor_wrapper.go

PR #83, opened 2022-05-18
Cleanup temp files

rbehjati:


build/build.go
```diff
@@ -55,6 +56,7 @@ func Build(buildFilePath, gitRootDir string) (*slsa.Provenance, error) {
 			return nil, fmt.Errorf("couldn't fetch sources from %s: %v", buildConfig.Repo, err)
 		}
 		log.Printf("Fetched the repo into %q. See %q for any error logs.", repoInfo.RepoRoot, repoInfo.Logs)
+		tempFolder = repoInfo.RepoRoot
```
tiziano88, line 59: Alternatively, you could add a method `repoInfo.Cleanup()` on this object, and call `defer` on it here. Especially since we still want the cleanup to happen even if the function does not run until the end, right? WDYT?

verify/verify.go
```diff
@@ -54,6 +54,7 @@ func (verifier *ReproducibleProvenanceVerifier) Verify(provenanceFilePath string
 		return fmt.Errorf("couldn't load BuildConfig from provenance: %v", err)
 	}
 
+	tempFolder := ""
```
tiziano88, line 57: What's the plan for deduplicating this method vs the other above? Is there an issue for that already?

tiziano88 (approved):

PR #88, opened 2022-05-31
Command line options

### LINKED ISSUE #54: Use Flag Library (Or Similar) to Parse Command Line Options in Authorization Logic Process
cc @rbehjati

aferr:
Fixes #54

### DIFF
--- a/experimental/auth-logic/main.go
+++ b/experimental/auth-logic/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"flag"
 	"log"
 	"os"
 )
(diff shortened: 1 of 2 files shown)


aferr:
At the moment this PR depends on #86, but if you want to review this before the #86 is rebased and merged you can look at just the last commit.

experimental/auth-logic/main.go
```diff
@@ -26,11 +27,13 @@ import (
 // compiler can then run on the generated code.
 func main() {
 
-	appName := os.Args[1]
-	endorsementFilePath := os.Args[2]
-	provenanceFilePath := os.Args[3]
-	outputFilePath := os.Args[4]
-	queryName := os.Args[5]
+	appName := flag.String("app-name", "", "set name of application to be released")
```
rbehjati, line 30:
```suggestion
	appName := flag.String("app_name", "", "name of application to be released")
```

Based on what we have in the rest of the code, but I am not sure what is the more general convention.

rbehjati (approved):
Thanks :)

experimental/auth-logic/main.go
aferr: OK it looks like you are right that underscores should be used according to an internal style guide.

Issue #97, opened 2022-06-20
Add trusted builder key to Endorsement Release Authorization Logic Policy


aferr:
From what @rbehjati tells me, the outputs of trusted builders like Github Actions and Google Cloud Build will be signed using a unique keypair for these trusted builders. This signing does not yet exist (even independently of authorization logic), but we should add the relevant keys into the authorization logic policy once the signing is set up.

I think I heard that @mariaschett is working on adding key management for this (?).

Do either of you have more relevant details for this?

rbehjati:
We have this issue on Oak for tracking the signing of the Oak provenances: https://github.com/project-oak/oak/issues/2626, but we are still working out the details.

Issue #95, opened 2022-06-15
Generate endorsement file if authorization logic query passes


aferr:
We need a way to generate an endorsement file with the right contents if an authorization logic policy query is true (or generate an error if it is not true). We will use this to check if a predicate in the policy for releasing an endorsement file can be proved (is true), and generate the endorsement file if so.

rbehjati:
We already have internal tooling for this. Somewhere along the path, we will have to sync the repos.

aferr:
You already have tooling that *runs a query using the authorization logic compiler*, and if it passes, generates an endorsement file? It's just not on github? Can you send me a pointer to this in the appropriate channel?

rbehjati:
We have code that generates the endorsement file, no auth logic involved. The existing code is not on Github.

rbehjati:
For more clarity: We already have some verification logic that runs before generating the endorsement file. The idea is to only replace the verification part with authorization logic, not the entire pipeline.

aferr:
Ah OK got it that makes more sense :)

Sounds like it should be helpful.

If this work package otherwise makes sense, the way I'll order it is probably going to be:
 - I read up more on github actions and google cloud build to understand enough about how to interact with these
 - I write the auth logic policy(/policies) for pulling in this stuff
 - write the wrappers
 - do the issue we are commenting on.

So you should have time to sync the repos before I'll be able to make use of that.

aferr:
Hey @rbehjati, I recall that you mentioned you planned to move some things around after the workshop submission. Was moving this verification logic from the internal code base to this github code base part of that?

rbehjati:
No, it was not. And I did not get the time to make the other refactoring either.

Issue #92, opened 2022-06-15
Write an authorization logic policy for releasing endorsements.


aferr:
This policy should gather evidence needed to decide if an endorsement file can be released. This evidence includes output from one or more trusted builders such as guithub actions and google cloud build.

## Here are some "product requirements":

Different projects might have different needs w.r.t the trusted builders. For example:
 - one team might only trust github actions, but not any other builder
 - another team might be OK with getting output from either builder
 - another team might insist on getting the same output from both builders
 - another team might want the same output from k out of n trusted builders
 - other teams might have different needs entirely

Most teams will not want to write authorization logic code at all because it's based on logic programming, and this is not a common skill set. We should minimize how much they have to interact with this.

## Proposed solution:

To meet the needs of different product teams and to reduce interaction with authorization logic, we will write a few "Policy Principals" that each represent a different policy for deciding if the builder output is acceptable. For example, one policy principal will specify "just Github actions is trusted", another will specify that "getting the expected output from either builder is fine", and so on. 

These policy principals are sort of like libraries -- they remove the need for product teams to specify these policies themselves. The policy for releasing an endorsement file will include a delegation (using the authorization logic features for delegation) to one of these policy principals. Most of the time, product teams will just change this delegation requiring just a small 1-2  LOC change to the endorsement generation policy. Of course, teams that have more specific needs and are willing to do so can choose not to delegate to one of these policy principals and write the entire policy themselves -- we expect that this will not be a common case though

## Some implementation details:
As is the case with the other wrappers, the policy principals and the policy for releasing an endorsement file will be generated using a go template so that variables can be filled in.

aferr:
This was fixed by #99 and #104

Issue #91, opened 2022-06-15
Authorization Logic for Endorsement File Release.


aferr:
At present, we have an authorization logic tool that runs the client-side verification process. Our tool for doing this produces an authorization logic policy that collects evidence for verifying if a previously released binary "is safe" to be consumed by the client.

In addition to this, we would like tooling that works similarly for parties _releasing_ a software binary. This tooling will be used by the product team releasing a binary instead of the client. It gathers the evidence needed to generate an endorsement file, includes a policy for deciding if the evidence gathered is sufficient, and uses this policy to decide if the release is acceptable. If the tool succeeds, an endorsement file is generated. If it fails, it throws an error.

This is a meta-issue for tracking the work to build the release tool to be used by the product team. Here are the sub-tasks:
 - [x] #92 
 - [x] #93
 - [ ] #95
 - [x] #94
 - [x] #98
 - [ ] #97

aferr:
@tiziano88 @rbehjati WDYT

PR #99, opened 2022-06-21
Top level and example endorsement release policy

aferr:
This is meant to implement #92.

This PR includes an executable binary for using authorization logic to decide if endorsement files can be produced. It takes as input:
 - One or more authorization logic files as input. These include a policy for releasing endorsement files written by the product team and zero or more other authorization logic files that might implement libraries.
 - A provenance file
It generates as output:
 - an authorization logic file that extends the input policies with the outputs of the provenance file wrapper and the unix time wrapper.

The build rule `endorsement_release_oak_functions` runs this binary with an example. The input files for this example are:
 - `experimental/auth-logic/endorsement-release/input_policy_examples/oak_endorsement_policy.auth_logic` -- the policy set by the Oak functions loader team for deciding when to release a binary
 - `experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic`
 The output real output is generated in:
`bazel-bin/experimental/auth-logic/endorsement-release/oak_endorsement_release_output_policy.auth_logic`
and a copy of this is stored in:
`experimental/auth-logic/endorsement-release/output_policy_examples/oak_endorsement_release_output_policy.auth_logic`

Note that there are clearly parts of this policy that need to be fixed because the rest of #91 is not implemented. For example
 - the keys of the trusted builder are not in the policy (as in #97 ) (waiting for signatures of trusted builders fully designed on the transparent release side)
 - an endorsement file does not get generated when the query runs (as in #95 )(we need to sync parts of this from google3)
 - The provenance file wrapper uses a principal name that depends on the name of the application. To be able to write fixed policy libraries that refer to outputs of the provenance file wrapper (as in github_actions_policy.auth_logic), we should just change the principal for the provenance wrapper to be a fixed one like `Provenance` instead of `oak_functions_loader-0x...::Provenance`.
 
I totally agree with this point @rbehjati made in #92
> I am not sure this solution would scale well. The example policies we discussed yesterday seemed quite simple in the end. It should be possible for most software engineers to write policies like those. But also, feel free to proceed with this design :) (or maybe I did not completely understand this part!)

Splitting this policy into multiple principals is definitely overkill for this use-case. An advantage of writing it in this more complicated way though is that it gives us a relatively simple way to demonstrate:
 - the value of authorization logic because it uses a delegation to a policy principal
 - the concept of policy principals which might have applications to other problems like: writing and agreeing to conform to a standard for certificate authorities, writing/conforming to identity policies (e.g., is this a cryptographer? is this a person with rust readability? is this an RFID reader with a particular specification?)
 and because this policy is otherwise simple, maybe this makes it useful for a blog post or other external communications.

Please do let me know if you would rather I get rid of the policy principals to make it simpler and fit in the one policy input provided by the oak functions loader team. (My opinion doesn't really matter here and I don't have a strong one -- @tiziano88 and @rbehjati can just decide).


experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic
```diff
@@ -0,0 +1,11 @@
+"GithubActionsPolicy" says {
```
rbehjati, line 1: Is this file auto-generated? I have seen repos that have a `generated` folder for auto-generated files. I think that is useful conventions that we could use too.

experimental/auth-logic/endorsement-release/BUILD
```diff
@@ -0,0 +1,58 @@
...
+  cmd = """
+  export PROVENANCE=$(location //experimental/auth-logic/test_data:oak_provenance.json)
+  export OAK_ENDORSEMENT_POLICY=$(location //experimental/auth-logic/endorsement-release/input_policy_examples:oak_endorsement_policy.auth_logic)
+  export GITHUB_ACTIONS_POLICY=$(location //experimental/auth-logic/endorsement-release/input_policy_examples:github_actions_policy.auth_logic)
+  ./$(location :endorsement_release_policy_generator) --app_name oak_functions_loader:0f2189703c57845e09d8ab89164a4041c0af0a62 --provenance $$PROVENANCE --auth_logic_inputs $$OAK_ENDORSEMENT_POLICY --auth_logic_inputs $$GITHUB_ACTIONS_POLICY --auth_logic_out $(RULEDIR)/oak_endorsement_release_output_policy.auth_logic
+  """,
+  tools = [
+    ":endorsement_release_policy_generator",
+  ]
+)
```
rbehjati, line 58: Newline?

experimental/auth-logic/endorsement-release/main.go
```diff
@@ -0,0 +1,80 @@
...
+func (someStringArray *stringArray) Set(value string) error {
+	*someStringArray = append(*someStringArray, value)
+	return nil
+}
+
+func main() {
+
+	var authLogicInputs stringArray
+
+	appName := flag.String("app_name", "", "set name of application to be released")
```
rbehjati, line 43:
```suggestion
	appName := flag.String("app_name", "", "name of application to be released")
```
Or do we use "set ..." everywhere?

experimental/auth-logic/endorsement-release/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"flag"
+	"log"
+	"os"
+	"strings"
+)
+
+// This program accepts multiple auth logic files as input by passing them
+// into an array variable. To allow a []string to be used in flag.var, it must
```
rbehjati, line 25:
```suggestion
// into an array variable. To allow a []string to be used in flag.Var, it must
```

experimental/auth-logic/endorsement-release/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+	appName := flag.String("app_name", "", "set name of application to be released")
+	provenanceFilePath := flag.String("provenance", "", "set path of provenance file")
+	outputAuthLogicFilePath := flag.String("auth_logic_out", *appName+"_endorsement_release_policy.auth_logic", "set path for generated output authorization logic verification code")
+	flag.Var(&authLogicInputs, "auth_logic_inputs", "set one or more auth logic input files; to provide multiple inputs, call the parameter more than once with each input, for example, `--auth_logic_inputs file1 --auth_logic_inputs file2`")
+
+	flag.Parse()
+
+	// Part of the code for building a project using provenance
+	// files changes the working directory. This binary needs to keep
```
rbehjati, line 51: Really? Which part changes the working directory?

experimental/auth-logic/endorsement-release/top_level.go
```diff
@@ -0,0 +1,79 @@
...
+	".decl attribute hasPublicKey(hash : Sha256Hash)\n"
+
+// generateEndorsenent takes an application name, the path to an endorsement
+// file for that application, the path to a provenance file for that
+// application, and emits authorization logic code (as a string) that runs the
+// transparent release verification process.
+func generateEndorsement(authLogicInputs []string, appName, provenanceFilePath string) (string, error) {
+
+	var authLogicFileContents = ""
+	for _, authLogicInputFile := range authLogicInputs {
```
rbehjati, line 37: I don't quite understand this part. Is it the input policy?

rbehjati (approved):
Sorry for the late review.

> Please do let me know if you would rather I get rid of the policy principals to make it simpler and fit in the one policy input provided by the oak functions loader team.

I don't have a strong opinion about it. Let's keep them for now, and we'll reevaluate later if we want to keep them or not. 

I think it would be nice to add a README here and/or for `experimental/auth-logic`, with some of the information you have in the description of the PR. Feel free to add this in a separate PR.

experimental/auth-logic/endorsement-release/main.go
aferr: This comment is just referring to #60. The comment is just directly copied from the clinet-verification code.

rbehjati: Is this part of the code even related to #60? Does the code here call `FetchSourcesFromRepo`? I think it should not.

aferr: I think it does indirectly... let me check more thoroughly later...

experimental/auth-logic/endorsement-release/top_level.go
aferr: This function combines the input policy with the outputs from the wrappers. The function can take any number of authorization logic policy files as inputs, so the for loop that this comment is about just concatenates all the authorization logic input files before concatenating them with the outputs from the provenance wrappers.

It's potentially useful to include multiple policy files as inputs because then some policy files can be shared across applications. In the example that generates the output that's checked in, there are two inputs:
 - `github_actions_policy.auth_logic`
 - `oak_endorsement_policy.auth_logic`

(With a slight change to the authorization logic compiler, we can just pass the outputs from the wrappers and these input policies directly to the authorization logic compiler. The only reason we can't already do this is that we need a tweak avoid collisions with relation declarations. I haven't fixed this yet because the authoriation logic compiler is being reimplemented in C++).

aferr:
@tiziano88 the generic policy that I think you're looking for is split between the two input files:
`experimental/auth-logic/endorsement-release/input_policy_examples/oak_endorsement_policy.auth_logic` and
`experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic`

Is this what you're looking for?

experimental/auth-logic/endorsement-release/top_level.go
rbehjati: Sounds good.

experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic
```diff
@@ -0,0 +1,11 @@
...
+    "Provenance" canSay any_app
+        has_expected_hash_from(any_hash, "Provenance").
```
tiziano88, line 3: Not sure I understand the intuitive meaning of `has_expected_hash_from` (i.e. why is the second argument always "Provenance"?); in general, how do we document such predicates?

experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic
```diff
@@ -0,0 +1,11 @@
...
+
+    "Provenance" canSay any_app
+        has_builder(any_builder).
+
+    BuildPolicyAllowRelease(any_app, any_hash) :-
```
tiziano88, line 8: What is the difference between predicates in SnakeCase vs lower_underscore  format? Should we just pick one?

experimental/auth-logic/endorsement-release/input_policy_examples/oak_endorsement_policy.auth_logic
```diff
@@ -0,0 +1,40 @@
...
+    // The product team can change this line to pick how to accept outputs from 
+    // trusted builders. For example:
+    // - one team might only trust github actions, but not any other builder
+    // - another team might be OK with getting output from either builder
+    // - another team might insist on getting the same output from both builders
+    // - another team might want the same output from k out of n trusted builders
+    // Product teams will usually not want to write much authorization logic. 
+    // So we offer policy principals that encode these common options. The 
+    // product team can pick one of these options by _delegating_ to principals
+    // that encodes one of these policy choices. Product teams with different needs
```
tiziano88, line 11:
```suggestion
    // that encode one of these policy choices. Product teams with different needs
```

experimental/auth-logic/endorsement-release/input_policy_examples/oak_endorsement_policy.auth_logic
```diff
@@ -0,0 +1,40 @@
...
+    // can also choose not to delegate to one of these policies and write a complete
+    // policy on their own
+
+    "GithubActionsPolicy" canSay BuildPolicyAllowRelease("oak_functions_loader:0f2189703c57845e09d8ab89164a4041c0af0a62::Binary", any_hash).
```
tiziano88, line 15: Eventually I don't even think we need the entire `oak_functions_loader:0f2189703c57845e09d8ab89164a4041c0af0a62::Binary` thing, in principle just `sha256:0f2189703c57845e09d8ab89164a4041c0af0a62` should be sufficient, right?

experimental/auth-logic/endorsement-release/top_level.go
```diff
@@ -0,0 +1,79 @@
...
+
+const relationDeclarations = ".decl BuildPolicyAllowRelease(binary : Principal, hash : Sha256Hash)\n" +
+	".decl RealTimeNsecIs(time : Number)\n" +
+	".decl attribute hasPublicKey(hash : Sha256Hash)\n"
+
+// generateEndorsenent takes an application name, the path to an endorsement
+// file for that application, the path to a provenance file for that
+// application, and emits authorization logic code (as a string) that runs the
+// transparent release verification process.
+func generateEndorsement(authLogicInputs []string, appName, provenanceFilePath string) (string, error) {
```
tiziano88, line 34: Based on the doc comment, the output of this function is not actually an endorsement (though it wil be used to decide whether or not to generate an endorsement file, which will be done out of band), so I think the naming is not quite accurate.

experimental/auth-logic/endorsement-release/README.md
```diff
@@ -0,0 +1,22 @@
...
+ - An authorization logic file that extends the input policies with the outputs of the provenance file wrapper and the unix time wrapper.
+
+The build rule endorsement_release_oak_functions runs this binary with an example. The input files for this example are:
+ - `experimental/auth-logic/endorsement-release/input_policy_examples/oak_endorsement_policy.auth_logic` -- the policy set by the Oak functions loader team for deciding when to release a binary
+ - `experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic`
+
+The real output is generated in:
+ - `bazel-bin/experimental/auth-logic/endorsement-release/oak_endorsement_release_output_policy.auth_logic`
+and a copy of this is stored in:
+ - `experimental/auth-logic/endorsement-release/output_policy_examples/oak_endorsement_release_output_policy.auth_logic`
```
tiziano88, line 18: Is there anything that automates / checks that this copy is correct?

tiziano88 (approved):
Thanks, I think we are on the same page, and the policy looks more like what I had in mind, i.e. something that applies to a collection of released binaries. I like the ability to provide multiple files as input too, though I think when we start talking about the policy externally we would probably join them together first anyways, so I wonder whether we should just assume that it is done offline before invoking this tool. Anyways that would be very easy to change later, we can keep thinking about it and see what feels more ergonomic later on.

experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic
aferr: Here's an explanation: [link](https://github.com/project-oak/transparent-release/blob/0048299cf2cda06bd7b79eac0ed75a58ffb159d8/experimental/auth-logic/wrappers/verifier_wrapper.go#L34).

Authorzation logic is essentially being re-implemented in C++ as part of Raksha. This prototype is still using the Rust version. The new version of authorization logic for Raksha will also be iterated on, and it will fix this by allowing a restricted form of "says" on the RHS that should still be decidable. This is the workaround for now, but this can be fixed later once we migrate to the C++ version.

experimental/auth-logic/endorsement-release/README.md
aferr: No, but I think now is not the time to do this, because now that I think you're on board with how this works, I will probably adjust the example policy somewhat. Here's an issue: #103

experimental/auth-logic/endorsement-release/input_policy_examples/github_actions_policy.auth_logic
aferr: If we really intend to do a blog post or some other external communication of this before migrating to the C++ version, I can make this change in the Rust version in time for that external comm, of course.

experimental/auth-logic/endorsement-release/input_policy_examples/oak_endorsement_policy.auth_logic
aferr: thanks fixed

experimental/auth-logic/endorsement-release/top_level.go
aferr: How about `verifyRelease` as the name. I made other fixes to the comment as well.

experimental/auth-logic/endorsement-release/README.md
aferr: I guess "copying" is not stalled or affected by changes to the policies (even though "checking" would be). In any case, this seems like a task for a separate PR to me.

PR #101, opened 2022-06-27
Add Builder.Id to provenance schema

### LINKED ISSUE #98: Extend provenance file wrapper to read the builder ID
The authorizaiton logic policies will depend on which trusted builder produced the provenance file. The builder ID shows up under the `builder` predicate [here](https://slsa.dev/provenance/v0.2). The other tooling around transparent release does not support this field yet, so this issue depends on adding support for getting this builder field from provenance files.

(information provided by @rbehjati )

aferr:
This also adds a statement about the builder ID to the provenance wrapper. It's likely that I'll adjust the statement produced by the provenance wrapper after #99 is merged.

This fixes #98


rbehjati (approved):
Nice. Thanks.

PR #105, opened 2022-06-29
Bump auth logic version

aferr:
This was needed to change the grammar for constants so that we can have newlines and other characters that are used in public keys.

This is related to [Raksha #593](https://github.com/google-research/raksha/pull/593)


aferr:
The important parts of this PR (the url, the hash) are also likely to be adjusted before merging, but this is useful so that I can work on the other in-flight PR

rbehjati (approved):

PR #110, opened 2022-07-04
Add structs for amber claims

### LINKED ISSUE #31: Implementation of endorsement structs in Go
In [the schema](https://github.com/project-oak/transparent-release/blob/main/schema/amber-endorsement/v1/statement.json), we currently have the time fields as strings, in the Go implementation we could use `time.Time` for these fields. This may require implementing a custom serializer, but would allow time comparisons directly on this struct. 

See the CustomMarshalJSON example in [https://pkg.go.dev/encoding/json](https://www.google.com/url?sa=D&q=https%3A%2F%2Fpkg.go.dev%2Fencoding%2Fjson)

Related issue: #30.

rbehjati:
* Adds structs for claims. This introduces v2 endorsements, and retires the earlier endorsement format. 
* Adds a function for parsing a v2 endorsement file into an intoto Statement with a ClaimPredicate.
* Adds an example claim (with AmberEndorsementV2 as the claimType) and a test to check the endorsement parsing.
* Moves `amber` to `pkg` as a step towards having the [recommended layout](https://github.com/golang-standards/project-layout). 

The main changes are in `pkg/amber`. The rest of the changes are just updated paths due to moving `amber` to `pkg` and moving a const to `amber`.

Also fixes #31.

@aferr This new format should eventually be used in Authorization Logic and replace the existing endorsements.


pkg/amber/endorsement.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+)
+
+// AmberEndorsementV2 is the ClaimType for Amber Endorsements V2. This is
+// expected to be used together with the AmberClaimV1 as the predicate type in
```
aferr, line 26: Just checking: here "predicate type" means the same thing as "ClaimType", right?

aferr (approved):

Issue #84, opened 2022-05-24
Add builder, invocation and metadata to generated SLSA provenances


rbehjati:
The [SLSA provenance v0.2 format](https://slsa.dev/provenance/v0.2#schema) has a few optional fields, in particular `invocation` and `metadata`, that we currently don't fill out in the provenances that we generate. 
To be able to fill these fields out when generating the provenances using `cmd/build`, the input `.toml` should be adjusted to provide `invocation`-related information. 

In particular, in case of [this GitHub workflow](https://github.com/project-oak/oak/blob/main/.github/workflows/provenance.yaml) the `invocation` could be filled out as follows: 

```json
"invocation.configSource.uri": "https://github.com/project-oak/oak/blob/f8c4c96375ca1d87a74238fc4450a81703d1d147/scripts/generate_provenance"
"invocation.configSource.digest": "<cryptographic digests of the file above>"
"invocation.configSource.entryPoint": ""
"invocation.parameters": "GitHub" 
"invocation.environment": "<info about the GitHub Actions runner>"

```
More generally, `invocation.configSource.entryPoint`, `invocation.parameters` and `invocation.environment` are optional and can be omitted. 

The `metadata` should be set as follows:
* `metadata.buildStartedOn` = The timestamp of when the build started, filled out by `cmd/build`.
* `metadata.buildFinishedOn` = The timestamp of when the build completed, filled out by `cmd/build`.
* `metadata.completeness` and `metadata.reproducible`: For now we don't fill these out.

The `builder.id` is another field that should be added as a command line flag, instead of being included in the toml file.

PR #111, opened 2022-07-04
Add invocation to  BuildConfig

rbehjati:
Fixes ##84

* Adds invocation details to the BuildConfig 
* Adds a new example config TOML file (build.toml)
* Updates the provenance generation logic

- [ ] Tests pass


testdata/build.toml
```diff
@@ -1,6 +1,17 @@
 repo = "https://github.com/project-oak/oak"
```
tiziano88, line 1: To confirm: this is the "input" TOML file that is used to determine what to build, right? e.g. https://github.com/project-oak/oak/blob/main/buildconfigs/oak_functions_loader_base.toml . If so, I am a bit confused by this change, I wouldn't expect additional fields to be added there (only in the generated provenance file).

rbehjati: This is not exactly the same file as https://github.com/project-oak/oak/blob/main/buildconfigs/oak_functions_loader_base.toml, but the one that is generated by the script: https://github.com/project-oak/oak/blob/ab106d1129fc469c36b2d37fcadd617aea1f4193/scripts/generate_provenance#L55-L68

We need this additional info to be able to create a complete provenance file.

> I wouldn't expect additional fields to be added there (only in the generated provenance file).

Where would the provenance file generator get this information from?

tiziano88: As long as we don't expect to add these fields to the actual build files (e.g. https://github.com/project-oak/oak/blob/main/buildconfigs/oak_functions_loader_base.toml ) then that's fine, that's the only thing I wanted to check at the moment.

Perhaps we should think more about the naming though, this file does look very similar to the build one I was referring to (and the left side of the diff contains exactly the same information as that). How should we call them when referring to them informally? build.toml and provenance.toml? something else?

rbehjati: > As long as we don't expect to add these fields to the actual build files (e.g. https://github.com/project-oak/oak/blob/main/buildconfigs/oak_functions_loader_base.toml ) then that's fine, that's the only thing I wanted to check at the moment.

I cannot guarantee that, but hopefully we won't have to add many fields to it. I'd be happy to have a chat :) 

>Perhaps we should think more about the naming though, this file does look very similar to the build one I was referring to 

The TOML files that we have on the Oak repo are static. We could call them static or partial build files. 

> (and the left side of the diff contains exactly the same information as that)

That is not entirely true. In the build file here, we have additional fields that cannot be static. In particular the `commit_hash` and the `builder_image`.

> How should we call them when referring to them informally? build.toml and provenance.toml? something else?

We could call the build file here provenance since it no longer contain only the `BuildConfig`, but also the `invocation` details. Or perhaps invocation details could be provided in a separate file. WDYT?

tiziano88: Thanks, perhaps we should call the existing one "input build file"? "static build file" works too, as you suggested. But we should make sure that this file can indeed be left mostly untouched in the repository over time. And it would be good to have a dedicated minimal Go / JSON schema for it (which I think includes the fields that are already there at the moment, and not much more). In fact probably even the repo field could be removed from there at some point (since it's always meant to be found inside an existing repository, and in principle it should work even if the thing to build is not in a git repository). Also the `expected_binary_sha256_hash` should probably go away too at some point, since IIRC we are going to rely on comparing the generated provenance files (and their subject fields specifically) in order to confirm whether a binary can be built reproducibly.

But I'm not sure why we need the extra information to also end up in another TOML file; I understand why we need the information in order to generate the provenance file, but shouldn't that just internal to the provenance generation process, instead of being written to another TOML file first? And eventually it ends up being part of the generated provenance file as JSON, right?

rbehjati: > Thanks, perhaps we should call the existing one "input build file"? "static build file" works too, as you suggested.

Either one works for me. But this a name that we use informally. We don't have it documented anywhere (yet), right?

> And it would be good to have a dedicated minimal Go / JSON schema for it (which I think includes the fields that are already there at the moment, and not much more).

That only makes sense to me if there was some logic somewhere to verify those static build files in isolation. Once we have that logic implemented somewhere, we can add the schema too. 

> In fact probably even the repo field could be removed from there at some point (since it's always meant to be found inside an existing repository, and in principle it should work even if the thing to build is not in a git repository).

Possibly. This in my opinion is a minor improvement that can be done later. 

> Also the `expected_binary_sha256_hash` should probably go away too at some point, since IIRC we are going to rely on comparing the generated provenance files (and their subject fields specifically) in order to confirm whether a binary can be built reproducibly.

I agree. Will fix it as part of this PR. 

> But I'm not sure why we need the extra information to also end up in another TOML file; I understand why we need the information in order to generate the provenance file, but shouldn't that just internal to the provenance generation process, instead of being written to another TOML file first?

The provenance generation process is done by the `cmd/build` tool here. This tool has no knowledge of the invocation information or the environment, the builder (e.g., GitHub Actions), or the script (e.g., [Oak's provenance_generator script](https://github.com/project-oak/oak/blob/ab106d1129fc469c36b2d37fcadd617aea1f4193/scripts/generate_provenance)) that it was called from. I cannot think of any way to include this information in the provenance file other than providing them as input to `cmd/build` somehow. 

> And eventually it ends up being part of the generated provenance file as JSON, right?

That is correct.

tiziano88: I think we currently need to serialize this additional data as TOML because we need to pass it from a bash script to the build tool (and TOML is convenient for that, even though we are not actually generating TOML properly, but an approximation of it based on string concatenation in bash), but I was assuming that we don't want to keep maintaining the bash script, and all the logic would eventually go into the build tool (at which point the build tool just reads the static build file, and then it builds directly the generated provenance JSON file, including the invocation details). Alternatively these could be command line flags to the build tool too. To be clear, it's also fine to keep using TOML, but I am trying to clarify who / what is expected to write / generate each version of these TOML files, since they seem to serve quite different purposes, despite having some fields in common. But let me know if there is a better way I should be thinking about it.

rbehjati: > but I was assuming that we don't want to keep maintaining the bash script

Which bash script are you referring to? Do you mean [this one](https://github.com/project-oak/oak/blob/ab106d1129fc469c36b2d37fcadd617aea1f4193/scripts/generate_provenance)? If so, this is not something that will change often. The reason for this PR and the related changes is that originally we were not aiming for full conformance with the SLSA provenance. This PR and the related issue is an attempt to get better conformance with SLSA. We should minimize the impact on the users of this tool, but some downstream changes are inevitable. 

> and all the logic would eventually go into the build tool (at which point the build tool just reads the static build file, and then it builds directly the generated provenance JSON file, including the invocation details).

That is not always possible. The build tool has to get the information it needs, in particular about the invocation details, from somewhere. Perhaps I am misunderstanding what Invocation is. Here is the description:

```
Identifies the event that kicked off the build. When combined with materials, 
this SHOULD fully describe the build, such that re-running this invocation results 
in bit-for-bit identical output (if the build is [reproducible](https://reproducible-builds.org/)).

 MAY be unset/null if unknown, but this is DISCOURAGED.
```

What do you think invocation should be in case of Oak binaries built on GitHub? 

> since they seem to serve quite different purposes, despite having some fields in common. But let me know if there is a better way I should be thinking about it.

What do you mean by different purposes?

rbehjati: @asraa Do you have any ideas? The issue is #84.

asraa: Sorry for the delay! Here are some thoughts:

> but shouldn't that just internal to the provenance generation process, instead of being written to another TOML file first? 

and 

>  all the logic would eventually go into the build tool (at which point the build tool just reads the static build file, and then it builds directly the generated provenance JSON file, including the invocation details). Alternatively these could be command line flags to the build tool too.

I think I agree here, these fields should be generated by the builder via some kind of inspection. From reading this PR, it seems that the intent right now is to add fields to the "intermediate"/modified TOML. But I am not sure where it is getting generated. Is it meant to get generated in this [step](https://github.com/project-oak/oak/blob/ab106d1129fc469c36b2d37fcadd617aea1f4193/scripts/generate_provenance#L64-L67) (and is just not added yet?)

If so, then why can't the build tool itself take these as inputs or env vars to expect? 

>  This tool has no knowledge of the invocation information or the environment, the builder (e.g., GitHub Actions), or the script (e.g., [Oak's provenance_generator script](https://github.com/project-oak/oak/blob/ab106d1129fc469c36b2d37fcadd617aea1f4193/scripts/generate_provenance)) that it was called from. I cannot think of any way to include this information in the provenance file other than providing them as input to cmd/build somehow.

Maybe you can consider a sub-command like "dry-run" that only creates unsigned, intermediate provenance to pass to the `./cmd/build` tool in the next step? I agree it must be provided as input in some way. If run inside a GitHub environment, it can detect the environment, or locally, maybe nothing.

Some other comments:
* I believe configSource should actually be repo and commit_hash specified anyway
* The entryPoint is actually the `generate_provenance` script, right? This is what is building the binary
* Params are the inputs to the scripts, I agree - again, since the script receives it, it can pass it in as input to the `./cmd/build`.

Please correct me if I'm misunderstanding!

rbehjati: Thanks for the comment Asra.

> If so, then why can't the build tool itself take these as inputs or env vars to expect?

Generally, we have been leaning towards bundling the input parameters in a config file, rather than having them as command line arguments. Config files provide a more flexible API. But the config file and the flags serve slightly different purposes. Moreover, the builder flags, and build configs come from different sources and change at different frequencies, so it does make sense to handle them separately. 

> Some other comments:
> 
> * I believe configSource should actually be repo and commit_hash specified anyway
> * The entryPoint is actually the `generate_provenance` script, right? This is what is building the binary
> * Params are the inputs to the scripts, I agree - again, since the script receives it, it can pass it in as input to the `./cmd/build`.

Our plan is to completely retire our own provenance format, and use SLSA's format. I think it is better to discuss the format in the context of https://github.com/slsa-framework/slsa-github-generator/issues/23.

rbehjati:
No longer relevant.

PR #118, opened 2022-08-03
Remove ExpectedBinarySha256Hash from BuildConfig

rbehjati:
With the recent changes in our design, there is no longer a need for `ExpectedBinarySha256Hash`. This PR removes that.

### DIFF
--- a/common/common_test.go
+++ b/common/common_test.go
@@ -30,7 +30,7 @@ const testdataPath = "../testdata/"
 const provenanceExamplePath = "schema/amber-slsa-buildtype/v1/example.json"
 
 func TestComputeBinarySha256Hash(t *testing.T) {
-	want := "56893dbba5667a305894b424c1fa58a0b51f994b117e62296fb6ee5986683856"
+	want := "3dbf6017c84f2a6be8d1d914ff6da2b9a34829b1846a342b8b73856fa53d4d6b"
 	path := filepath.Join(testdataPath, "build.toml")
 	got, err := computeSha256Hash(path)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


common/common_test.go
```diff
@@ -141,20 +139,19 @@ func TestGenerateProvenanceStatement(t *testing.T) {
 	assert("builderImage", predicate.Materials[0].URI, "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320")
 	assert("commitHash", predicate.Materials[0].Digest["sha256"], "53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320")
 	assert("subjectName", prov.Subject[0].Name, "build.toml-0f2189703c57845e09d8ab89164a4041c0af0a62")
-	assert("expectedSha256Hash", prov.Subject[0].Digest["sha256"], "56893dbba5667a305894b424c1fa58a0b51f994b117e62296fb6ee5986683856")
+	assert("expectedSha256Hash", prov.Subject[0].Digest["sha256"], "3dbf6017c84f2a6be8d1d914ff6da2b9a34829b1846a342b8b73856fa53d4d6b")
```
tiziano88, line 142: Shouldn't this be removed too?

tiziano88 (approved):
Thanks!

common/common_test.go
rbehjati: Changed the name. This just checks the digest of the subject.

PR #119, opened 2022-08-03
Make tests more stable

rbehjati:
Currently most tests depend on the exact values in the testdata/example files. As a result if we update an example file, many changes are required in the tests. In most cases, the exact values are not relevant for the test logic, but rather the presence or absence of values needs to be checked. This PR updates the tests to check invariants on the values (e.g., length of a SHA256 hash) rather than the exact values.

### DIFF
--- a/common/common_test.go
+++ b/common/common_test.go
@@ -131,26 +131,14 @@ func TestGenerateProvenanceStatement(t *testing.T) {
 	buildConfig := predicate.BuildConfig.(amber.BuildConfig)
 
 	// Check that the provenance is generated correctly
-	assertEq(t, "repoURL", predicate.Materials[1].URI, "https://github.com/project-oak/oak")
-	assertNonEmpty(t, "subjectName", prov.Subject[0].Name)
-	assertEq(t, "subjectDigest", len(prov.Subject[0].Digest["sha256"]), wantSHA256HexDigitLength)
-	assertEq(t, "commitHash length", len(predicate.Materials[1].Digest["sha1"]), wantSHA1HexDigitLength)
-	assertEq(t, "builderImageID length", len(predicate.Materials[0].Digest["sha256"]), wantSHA256HexDigitLength)
-	assertEq(t, "builderImageURI", predicate.Materials[0].URI, fmt.Sprintf("gcr.io/oak-ci/oak@sha256:%s", predicate.Materials[0].Digest["sha256"]))
-	assertNonEmpty(t, "command[0]", buildConfig.Command[0])
-	assertNonEmpty(t, "command[1]", buildConfig.Command[1])
-}
-
-func assertEq[T comparable](t *testing.T, name string, got, want T) {
-	if got != want {
-		t.Errorf("Unexpected %s: got %v, want %v", name, got, want)
-	}
-}
-
-func assertNonEmpty(t *testing.T, name, got string) {
-	if len(got) == 0 {
-		t.Errorf("Unexpected %s: non-empty string must be provided", name)
-	}
+	testutil.AssertEq(t, "repoURL", predicate.Materials[1].URI, "https://github.com/project-oak/oak")
+	testutil.AssertNonEmpty(t, "subjectName", prov.Subject[0].Name)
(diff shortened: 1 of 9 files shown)


common/common_test.go
```diff
@@ -20,24 +20,28 @@ import (
...
 func TestComputeBinarySha256Hash(t *testing.T) {
-	want := "3dbf6017c84f2a6be8d1d914ff6da2b9a34829b1846a342b8b73856fa53d4d6b"
 	path := filepath.Join(testdataPath, "build.toml")
 	got, err := computeSha256Hash(path)
 	if err != nil {
 		t.Fatalf("couldn't get SHA256 hash: %v", err)
 	}
-	if got != want {
-		t.Errorf("invalid commit hash: got %s, want %s", got, want)
+	if got != wantTomlHash {
```
tiziano88, line 43: What is a TOML hash? Why is it related to the commit hash?

common/common_test.go
```diff
@@ -20,24 +20,28 @@ import (
...
 	"github.com/project-oak/transparent-release/internal/testutil"
 	"github.com/project-oak/transparent-release/pkg/amber"
 )
 
-const testdataPath = "../testdata/"
-const provenanceExamplePath = "schema/amber-slsa-buildtype/v1/example.json"
+const (
+	testdataPath             = "../testdata/"
+	provenanceExamplePath    = "schema/amber-slsa-buildtype/v1/example.json"
+	wantTomlHash             = "f8b8b42df90a32f24822a5cf527574598a7812209455cf293bf99b84de90934b"
```
tiziano88, line 31: I think most initialisms are incorrectly capitalised in this PR. See http://go/go-style/decisions#initialisms

pkg/amber/provenance_test.go
```diff
@@ -35,29 +40,33 @@ func TestExampleProvenance(t *testing.T) {
...
+}
+
+func assertEq[T comparable](t *testing.T, name string, got, want T) {
+	if got != want {
+		t.Errorf("Unexpected %s: got %v, want %v", name, got, want)
+	}
+}
+
+func assertNonEmpty(t *testing.T, name, got string) {
+	if len(got) <= 0 {
```
tiziano88, line 69: `if got == ""` seems more idiomatic

tiziano88 (approved):

rbehjati:
Thanks for the review :)

common/common_test.go
```diff
@@ -20,24 +20,28 @@ import (
...
-	path := filepath.Join(testdataPath, "build.toml")
+	path := filepath.Join(testdataPath, "static.txt")
 	got, err := computeSha256Hash(path)
 	if err != nil {
 		t.Fatalf("couldn't get SHA256 hash: %v", err)
 	}
-	if got != want {
-		t.Errorf("invalid commit hash: got %s, want %s", got, want)
+	if got != wantTOMLHash {
+		t.Errorf("invalid commit hash: got %s, want %s", got, wantTOMLHash)
```
tiziano88, line 44: I'm still not sure why this is a commit hash though.

common/common_test.go
```diff
@@ -70,23 +73,24 @@ func TestLoadBuildConfigFromProvenance(t *testing.T) {
...
 		t.Fatalf("couldn't load BuildConfig from provenance: %v", err)
 	}
+
 	checkBuildConfig(config, t)
 }
 
-func TestParseBuilderImageUriValidURI(t *testing.T) {
-	imageURI := "gcr.io/oak-ci/oak@sha256:53ca44b5889e2265c3ae9e542d7097b7de12ea4c6a33785da8478c7333b9a320"
-	alg, digest, err := parseBuilderImageURI(imageURI)
+func TestParseBuilderImageUriValidUri(t *testing.T) {
```
tiziano88, line 80:
```suggestion
func TestParseBuilderImageURIValid(t *testing.T) {
```

tiziano88 (approved):

common/common_test.go
rbehjati: Ah! Good catch! I had not noticed it!

PR #121, opened 2022-08-19
Move packages implementing internal behaviour to internal

rbehjati:
* Moves common to internal
* Renames build to builder and moves it to internal
* Renames verify to verifier and moves it to internal
* Renames cmd/build to cmd/builder
* Renames cmd/verify to cmd/verifier

`build` is renamed to `builder` because a "build" folder is treaded in a special way by some tools (e.g., bazel). To be consistent `verify` is also renamed to `verifier`.


mariaschett (approved):
Thank you! I think the renaming is a very good idea!

PR #122, opened 2022-08-22
Enable gosec linting

### LINKED ISSUE #117: Enable gosec and tagliatelle linting rules
#116 replaces `golint` with `golangci-lint`. In the configuration included in #116 for `golangci-lint` has disabled some rules, because the code was not conforming to those. Ideally we should enable at least `gosec` and `tagliatelle` and fix the code accordingly.

rbehjati:
Fixes #117

This PR
* Enables `gosec` linter as a result of which the permission for writing the provenance file is changed from `0644` to `0600`. 
* Enables the `tagliatelle` linter.
* Removes deprecated linters from the linting config in golangci-linters.yaml. 

- [x] Tests pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
           go-version: "=1.18.4"
       - name: Run golint
         run: |
-          go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.47.2
+          go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.48.0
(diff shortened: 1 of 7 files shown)


mariaschett (approved):
Looks good!

Now only the user can read and write the file, but group and other cannot any more. I can't see any problem here, but I am not an expert on permissions.

rbehjati:
> Looks good!
> 
> Now only the user can read and write the file, but group and other cannot any more. I can't see any problem here, but I am not an expert on permissions.

Perfect. Thanks for the review :)

PR #123, opened 2022-08-30
Add environment variables HOST_UID and HOST_GID to docker run

mariaschett:
- [x] Tests pass

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -165,6 +165,13 @@ func (b *BuildConfig) Build() error {
 		"--workdir=/workspace",
 		// Remove the container file system after the container exits.
 		"--rm",
+		// The host uid and gid are needed to give the host access rights.
+		// --env has to be in seperate string, otherwise docker 
+		// interprets "--env HOST_UID=%d" as single flag instead of flag with
+		// argument. With -e docker interprets "HOST_UID=%d" as one environment
+		// variable.
+		"--env", fmt.Sprintf("HOST_UID=%d", os.Getuid()),
+		"--env", fmt.Sprintf("HOST_GID=%d", os.Getgid()),
 		// Get a pseudo-tty to the docker container.
 		// TODO(razieh): We probably don't need it for presubmit.
 		"--tty"}
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -166,7 +166,7 @@ func (b *BuildConfig) Build() error {
 		// Remove the container file system after the container exits.
 		"--rm",
 		// The host uid and gid are needed to give the host access rights.
-		// --env has to be in seperate string, otherwise docker 
+		// --env has to be in seperate string, otherwise docker
 		// interprets "--env HOST_UID=%d" as single flag instead of flag with
 		// argument. With -e docker interprets "HOST_UID=%d" as one environment
 		// variable.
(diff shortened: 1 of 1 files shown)


internal/common/common.go
```diff
@@ -165,6 +165,13 @@ func (b *BuildConfig) Build() error {
 		"--workdir=/workspace",
 		// Remove the container file system after the container exits.
 		"--rm",
+		// The host uid and gid are needed to give the host access rights.
+		// --env has to be in separate string, otherwise docker
+		// interprets "--env HOST_UID=%d" as single flag instead of flag with
```
tiziano88, line 170: With the whitespace it makes sense that it would not work as it would be parsed as a single flag, but I think it should work with `=` (i.e. `--env=HOST_UID=%d`)?

internal/common/common.go
```diff
@@ -165,6 +165,13 @@ func (b *BuildConfig) Build() error {
...
+		// argument. With -e docker interprets "HOST_UID=%d" as one environment
+		// variable.
+		"--env", fmt.Sprintf("HOST_UID=%d", os.Getuid()),
```
tiziano88, line 173: As a wider discussion, we should think whether this is really what we want. Since we are passing the host UID and GID at runtime, in principle the build may do something completely different based on those values (e.g. insert a backdoor in the compiled binary if UID=123). I wonder if instead we should just run everything as root, and `docker cp` the resulting artifact(s) out of the resulting image (but not sure exactly how to do that).

tiziano88 (approved):

internal/common/common.go
mariaschett: Might be. Do you think it is worth a try?

internal/common/common.go
mariaschett: So my conclusion was the only way to get the artifacts out of docker (with user permissions) is to use HOST_UID/GID, e.g., [1]. 

How would "insert a backdoor if UID=123" work? Where would you write that?

[1] https://jtreminio.com/blog/running-docker-containers-as-current-host-user/#ok-so-what-actually-works

tiziano88: The only reason why we currently map the UID:GID is that we run this interactively during development, and we expect the intermediate artifacts to be used by the developer (e.g. build a binary and run it locally, or generate some JSON and edit it), which only works (nicely) if everything is owned by the same user. But for a release process, the only output is a binary, and its ownership is not really part of the identity of that binary (i.e. the hash of the binary does not include who owns it -- it's the same binary whether it's owned by root or by another user).

I think we should have that discussion separately though; for instance, another option would be to take the base image and `COPY`/`ADD` the workspace directory to it, then build the resulting image (instead of running it), and `docker cp` the artifact out of that image. Then the release process is guaranteed to be independent of the UID.

Inserting a backdoor for some UID currently would be just a matter of checking the UID env variable, and doing something different based on it (e.g. in xtask, or in any compiler that is used by it, or even in some scripts).

tiziano88 (approved):

rbehjati (commented):
Thanks for looking into this. I have a few questions.

- What are the default values for HOST_UID and HOST_GID?
- How does this change affect reproducibility? We want the build to generate the same binary regardless of the system on which it is run. Does this change guarantee that? 
- I think this needs to be be described in the readme, so that the user knows that internally the builder sets HOST_UID and HOST_GID. WDYT?

Issue #113, opened 2022-07-20
Generate test coverage reports


rbehjati:
A test coverage report can be generated using `go test -cover` (see https://tip.golang.org/doc/go1.2#cover). However, we cannot use it directly on the top-level repo directory, since we cannot run the authorization logic tests without bazel. 

A test coverage report can be generated with bazel (https://docs.bazel.build/versions/main/coverage.html). But I was not able to generate html reports using `genhtml` on output generated by the `bazel coverage --combined_report=lcov [target]` command.

PR #124, opened 2022-09-07
Check the hash of the binary in the verifier

rbehjati:
The modification in #118 that removed `ExpectedBinarySha256Hash` from `BuildConfig` mistakenly also removed the verification logic related to checking the hash of the binary in the verifier. This PR fixes that. 

This change adds a slightly modified version of the `VerifyBinarySha256Hash` logic back, and:
- Adds negative tests to be able to avoid these bugs in the future; additional test files are added to allow this.
- Replaces the Oak build example with a trivial example in testdata. This is to allow writing more tests (including negative tests) without making the CI too slow. In fact this PR reduces the CI time from more than 10 minutes to less than 2 minutes. For this:
  - The `testdata/build.toml` is renamed to `testdata/oak_build.toml` and a new `testdata/build.toml` is added with the simpler example,
  - A new json file, matching the new `testdata/build.toml` is added
- Adds a CI step for generating test coverage reports (ref #113, but does not completely fix that issue since the report is not complete and not easy to navigate)

Note that there is a permission issue when running `go test ./internal/verifier` with the new `build.toml` file. The same issue exists when running `bazel test` and `bazel run` with this example, so the README is updated to use `testdata/oak_build.toml` as the example. Also see #123 on the permission issue.


internal/verifier/verifier_test.go
```diff
@@ -34,11 +37,47 @@ func TestReproducibleProvenanceVerifier(t *testing.T) {
...
+	verifier := ReproducibleProvenanceVerifier{}
+
+	want := "failed to verify the hash of the built binary"
+
+	if err := verifier.Verify(invalidHashProvenancePath); !strings.Contains(err.Error(), want) {
+		t.Fatalf("want error containing message %q, got %v", want, err)
+	}
+}
+
+func TestReproducibleProvenanceVerifier_badCommand(t *testing.T) {
```
mariaschett, line 63: Really cool test! As idea: If we need the `bad_command_provenance.json` only in this test, shall we maybe hard code it as string in the test? Might also be faster to see the whole test then.

Similar with the `invalidHashProvenancePath`?

mariaschett (approved):
Thank you!

internal/common/common.go
```diff
@@ -318,6 +318,23 @@ func (b *BuildConfig) ChangeDirToGitRoot(gitRootDir string) (*RepoCheckoutInfo,
...
+// BuildConfig, and checks that this hash is equal to the given `expectedSha256Hash`.
+// Returns an error if the hashes are not equal.
+func (b *BuildConfig) VerifyBinarySha256Hash(expectedBinarySha256Hash string) error {
+	binarySha256Hash, err := b.ComputeBinarySha256Hash()
+	if err != nil {
+		return fmt.Errorf("couldn't get the hash of the binary: %v", err)
+	}
+
+	if binarySha256Hash != expectedBinarySha256Hash {
+		return fmt.Errorf("the hash of the generated binary does not match the expected SHA256 hash; got %s, want %v",
```
tiziano88, line 331: (nit) use %v or %s for both values

internal/verifier/verifier.go
```diff
@@ -67,6 +67,13 @@ func (verifier *ReproducibleProvenanceVerifier) Verify(provenanceFilePath string
 		return fmt.Errorf("couldn't build the binary: %v", err)
 	}
 
+	// The provenance is valid, therefore `expectedBinaryHash` is guaranteed to be non-empty.
+	expectedBinaryHash := provenance.Subject[0].Digest["sha256"]
+
```
tiziano88, line 72: I would still add explicit checks here, so that if the provenance happens to be malformed (e.g. the digest empty), we will get a nice error message. But if it's already guaranteed to be covered elsewhere, feel free to ignore (though I would still be a bit worried that over time it may get out of sync).

internal/verifier/verifier_test.go
```diff
@@ -34,11 +37,47 @@ func TestReproducibleProvenanceVerifier(t *testing.T) {
...
 
-	if err := verifier.Verify(schemaExamplePath); err != nil {
+	if err := verifier.Verify(validProvenancePath); err != nil {
 		t.Fatalf("couldn't verify the provenance file: %v", err)
 	}
 }
 
+func TestReproducibleProvenanceVerifier_invalidHash(t *testing.T) {
+	// The path to provenance is specified relative to the root of the repo, so we need to go one level up.
+	// Get the current directory before that to restore the path at the end of the test.
```
tiziano88, line 47: Why does the path need to be restored? Does it mean that it affects the execution of other test cases? That would seem problematic, since the execution of the tests would differ whether they are running concurrently or not. Is it possible to make the entire process not depend on the PWD and instead use absolute paths for everything? It would still be possible here to concat PWD with a relative path to obtain an absolute path and use that, but at least that would not mutate global state.

pkg/amber/provenance.go
```diff
@@ -89,6 +89,10 @@ func ParseProvenanceFile(path string) (*intoto.Statement, error) {
 		return nil, fmt.Errorf("could not unmarshal the provenance file:\n%v", err)
 	}
 
+	if len(statement.Subject) == 0 || statement.Subject[0].Digest["sha256"] == "" {
+		return nil, fmt.Errorf("at least one subject must be present, and it must have a sha256 digest")
```
tiziano88, line 93: What would be the semantic of multiple subjects? Perhaps we should limit the check to ensure exactly one subject for now, until we figure that out (if it's even something we want to allow at all -- I suspect not).

tiziano88 (approved):

internal/verifier/verifier_test.go
rbehjati: This is basically because when provenances are loaded, they are also validated against the schema. The path of the schema is hardcoded inside some other file, and getting the path right requires changing the directory. 

I have not been able to come up with a reliable and neat solution yet, but it certainly needs to be fixed in a separate PR. We have this bit of code in all tests, and it is causing a lot of clutter.

We plan to switch to the SLSA format, and retire the amber provenance format. So it probably is better to wait for that change. I believe this problem will automatically go away when that change is completed. Created #127.

tiziano88: +1 I think the basic functions (which are tested here, e.g. `verifier.Verify`) should deal with the actual content of the files (as strings), they should not assume they can read a file by its path. Then we can have helpers to just read the file at a path, and we can compose that by passing its content to the function in question.

PR #125, opened 2022-09-13
Remove old endorsement format

rbehjati:
The new endorsement format is based on the claims (see [example](https://github.com/project-oak/transparent-release/blob/9b5f98310dbbad675834474fa68c37d880687cb9/schema/amber-claim/v1/example.json)). There is no longer a need for the old format, and the endorser will use the new format too.


mariaschett (approved):
Thank you! Great!

PR #128, opened 2022-09-14
Add ednorser

rbehjati:
Fixes #<issue_number_goes_here>

- [ ] Tests pass


internal/endorser/endorser.go
```diff
@@ -0,0 +1,169 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	"github.com/project-oak/transparent-release/pkg/amber"
+)
+
+// GenerateEndorsement generates an endorsement statement for the given binary hash, for the given
```
tiziano88, line 33: (nit) I think in-toto and most other systems (also Docker and Ent) use the term "digest" for the noun, and "hash" only as verb. Up to you, but I think that may make things a bit more consistent.

internal/endorser/endorser.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	return &verifiedProvenances, nil
+}
+
+// verifyProvenances verifies that the given list of provenances are consistent and have the given
+// binaryHash as the subject. Provenances are consistent if they all have the same binary name and
+// binary hash. An error is returned if any of the conditions above are violated.
+// TODO(b/222440937): Document any additional checks.
+func verifyProvenances(provenances []amber.ValidatedProvenance, binaryHash string) error {
```
tiziano88, line 95: I'm not sure I fully understand why verifying provenances would be necessary for an endorsement.

In practice in most cases I would expect that before endorsing a binary, the provenance has already been checked (e.g. when checking in the file to google3), plus ideally more information about the quality of the binary; but just asserting that there exists any provenance for the binary seems a bit arbitrary, unless something else about the provenance is also checked. For example, anyone can produce a valid provenance of any binary, just by checking in to any github repository and running an "identity" builder that only emits that binary.

But probably you are thinking of something more interesting to check further down the line?

Or is the idea that as part of reviewing the binary for release, all the evidence considered is also forwarded to the endorsement, for future reference, in case someone wonders how the person endorsing the binary arrived to the conclusion that the binary was fine?

In any case, it does not hurt to have it here, but I'm worried about duplicating a subset of this logic, because that will also have to live somewhere else in order to ensure that binaries in google3 have the correct provenance (even if they never get endorsed). So perhaps a way of addressing my point is for this logic to be in the verifier instead of the endorser? In other words, we should not have to actually release a binary in order to find out that any of these preconditions related to provenance failed. Or is there another mechanism that will enforce that?

pkg/amber/endorsement.go
```diff
@@ -28,6 +30,23 @@ import (
 // version in `schema/amber-endorsement/v1`.
 const AmberEndorsementV2 = "https://github.com/project-oak/transparent-release/endorsement/v2"
 
+// VerifiedProvenanceSet encapsulates metadata about a non-empty list of verified provenances.
+type VerifiedProvenanceSet struct {
+	// Name of the binary that all validated provenances agree on.
+	BinaryName string
```
tiziano88, line 36: I didn't realise that the binary name has any significance; is it something that needs to be checked explicitly?

tiziano88 (approved):

pkg/amber/endorsement.go
mariaschett: +1.

mariaschett (approved):
LGTM, thanks!

pkg/amber/endorsement.go
rbehjati: I am not entirely sure if it is significant. If it has no significance, why is it included in the subject? This may be a question for in-toto actually.

If there are multiple provenances and they have different names, what should we do? In particular, which name should we use as the name in the subject of the generated endorsement? 

Finally, currently, in the Amber provenances, we include the Git commit hash in the name. This is a legacy decision, which will not be true for SLSA-generated provenances. So, perhaps the condition could be relaxed. 

Created a tracking issue: #144.

internal/endorser/endorser.go
rbehjati: > ...but just asserting that there exists any provenance for the binary seems a bit arbitrary...

That is not the idea. The goal is to make sure that provenances are valid before generating the endorsement. You mentioned that "the provenance has already been checked". How would the endorser know that? The ideas is that the endorser will itself run all the checks before generating the endorsement. 

> But probably you are thinking of something more interesting to check further down the line?

Yes. This is the endorser with the simplest set of checks. @mariaschett is designing a more realistic verifier.

> Or is the idea that as part of reviewing the binary for release, all the evidence considered is also forwarded to the endorsement, for future reference, in case someone wonders how the person endorsing the binary arrived to the conclusion that the binary was fine?

Yes. That is the idea, and all the input will be recorded as evidence in the endorsement statement. 

> In any case, it does not hurt to have it here, but I'm worried about duplicating a subset of this logic, because that will also have to live somewhere else in order to ensure that binaries in google3 have the correct provenance (even if they never get endorsed). So perhaps a way of addressing my point is for this logic to be in the verifier instead of the endorser? 

That is basically the plan. @mariaschett is designing a verifier **library** that will be used both here and in `cmd/verifier`. The `b/222440937` TODOs are exactly where the code should be replaced with calls to functionality from the verifier. 

> In other words, we should not have to actually release a binary in order to find out that any of these preconditions related to provenance failed. Or is there another mechanism that will enforce that?

The verifier tool can be used to perform any necessary checks without actually generating an endorsement statement. We could alternatively merge the two tools and add a dry-run option to the endorser to simulate the verifier. But I prefer having separate tools.

internal/endorser/endorser.go
rbehjati: Thanks. I fixed as many occurrences as I could without adding too many irrelevant changes to the PR.

rbehjati (commented):
Thanks for the reviews.

internal/endorser/endorser.go
tiziano88: > The verifier tool can be used to perform any necessary checks without actually generating an endorsement statement. We could alternatively merge the two tools and add a dry-run option to the endorser to simulate the verifier. But I prefer having separate tools.

Me too, that's why I was expecting this logic to be in the verifier, not in the endorser. But maybe the move to a verifier library you mentioned above will address that for me.

PR #136, opened 2022-10-03
Add instructions for rootless docker

rbehjati:


rbehjati:
@mariaschett Do you think this much documentation is enough? Or should I add more details, similar to the one that we have [in Oak](https://github.com/project-oak/oak/blob/ad09cdd71d60ab7e44aae10d6d889d911939b1f1/docs/development.md)?

mariaschett (approved):
Yes, I think that is sufficient documentation!

I don't think we want to keep this up-to-date, and I hope the people behind the link will :)

PR #139, opened 2022-10-06
Remove issuer and refactor metadata in Claim

### LINKED ISSUE #30: Explicity start and end validity fields in validityPeriod
https://github.com/project-oak/transparent-release/blob/4c5e62a0ccb01d8407335d2514dfb8835d936360/schema/amber-endorsement/v1/predicate.json#L8-L21

The `releaseTime` field seems specific to releases, but the validityPeriod object may well be used for arbitrary time-bounded trusted files, so I think the name is too specific. Also the time of a release may not correspond to the time where it should be trusted. e.g. a release may be made today, but it can only be trusted in the future (e.g. to leave time for reviewers to inspect it).

Additionally, I think we should always require the expiry time; the author of this object may use a time far in the future if necessary.

Finally ,I think it may be useful to use terminology from existing systems, like x509 certs, in which such fields are called NotBefore / NotAfter (see https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.notbefore?view=net-6.0)

### LINKED ISSUE #134: Issuer, signer, or both in the claim format?
The current design of the [claim format](https://github.com/project-oak/transparent-release/blob/50e7f8b68a58eb4f112790701570eab42199d6dc/docs/claim-transparency.md#the-claim-format) allows specifying an **issuer**. Currently we require the issuer to be the same as the signer. This brings a few questions and issues (see https://github.com/project-oak/transparent-release/pull/130#discussion_r978672866).

To fix this we have two options:
- Remove the issuer:
    - Rationale: if the issuer and the signer are always the same, having a separate field for the issuer is redundant. 
    - Does this design have any security drawbacks?
- Keep the issuer and make sure it is consistent with the signer 
    - In this case, to make sure the issuer and the singer match, the tool that generates the claim and fills out the issuer must also sign the claim. 
    - Currently the only tool that generates a claim is our endorser. Adding signing functionality to it can be complicated. Particularly, given that we will most likely need to support different signing technologies (e.g., on GitHub and GCB).
    - How is a signer identified? In our current design the issuer has an ID which is a URI. With both a signer and an issuer, how do we map the signer (often identified by its public key) to the URI representing the issuer? This is important when verifying the claims.

Note that in the [SLSA provenance format](https://slsa.dev/provenance/v0.2) the **builder** (which is a concept similar to issuer in our current design) is different from the signer. However, SLSA provenance specification states that: 
> “Consumers MUST accept only specific (signer, builder) pairs. For example, “GitHub” can sign provenance for the “GitHub Actions” builder, and “Google” can sign provenance for the “Google Cloud Build” builder, but “GitHub” cannot sign for the “Google Cloud Build” builder."

I think it is best to remove the issuer. In future, if we see a need for it, we can add it to the spec. 
What do you think @tiziano88, @mariaschett?

rbehjati:
Fixes #30 and fixes #134

### DIFF
--- a/docs/claim-transparency.md
+++ b/docs/claim-transparency.md
@@ -30,14 +30,11 @@ different types of claims.
 
   "predicateType": "https://github.com/project-oak/transparent-release/schema/claim/v1",
   "predicate": {
-    "issuer": {
-      "id": "<URI>"
-    },
     "claimType": "<URI>",
-    "metadata": {
-      "issuedOn": "<TIMESTAMP>",
-      "validFrom": "<TIMESTAMP>",
-      "expiresOn": "<TIMESTAMP>",
+    "issuedOn": "<TIMESTAMP>",
+    "validity": {
+      "notBefore": "<TIMESTAMP>",
+      "notAfter": "<TIMESTAMP>",
     },
     "claimSpec": { /* object */ },
     "evidence": [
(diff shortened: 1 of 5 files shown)


pkg/amber/claim.go
```diff
@@ -116,13 +104,21 @@ func validateClaimPredicate(predicate ClaimPredicate) (*ClaimPredicate, error) {
...
 
-	// Verify that ExpiresOn is greater than IssuedOn, if the former is provided.
-	if predicate.Metadata != nil {
-		if predicate.Metadata.ExpiresOn != nil &&
-			predicate.Metadata.ExpiresOn.Before(*predicate.Metadata.IssuedOn) {
-			return nil, fmt.Errorf("expiredOn (%v) is before issuedOn (%v)",
-				*predicate.Metadata.ExpiresOn,
-				*predicate.Metadata.IssuedOn)
+	// Verify that NotBefore is after than IssuedOn (inclusive).
+	if predicate.Validity != nil {
```
mariaschett, line 108: Can predicate.Validity be nil or would that be an error in itself? I think predicate.Validity is not optional, but maybe I am missing something.

Either way, I think you could check `predicate.Validity != nil` only once.

docs/claim-transparency.md
```diff
@@ -60,27 +57,19 @@ This section describes the semantics of each field in the claim format:
...
+- **validity** _(object, required)_:
+  Validity duration of the claim.
+
+  - **validity.notBefore** _(string ([Timestamp](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#Timestamp)), required)_:
+    The timestamp from which the claim is effective, and the artifact is endorsed for use. Must be
+    equal or after the issuedOn timestamp.
+  - **validity.notAfter** _(string ([Timestamp](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#Timestamp)), required)_:
     The timestamp of when the artifact is no longer endorsed for use. This, combined with the
-    validFrom field, is particularly useful for implementing passive revocation.
+    `notBefore` field, is particularly useful for implementing passive revocation.
```
mariaschett, line 72: You could move "This, combined with `notBefore` field, is particularly useful for implementing passive revocation." to up **validity**, i.e., "Validity duration of a claim enables implementing passive revocation."

pkg/amber/claim.go
```diff
@@ -116,13 +104,21 @@ func validateClaimPredicate(predicate ClaimPredicate) (*ClaimPredicate, error) {
...
+		if predicate.Validity.NotBefore.Before(*predicate.IssuedOn) {
```
mariaschett, line 109: Do we want to add tests for those two error cases, too?

mariaschett (approved):
Looks great!

rbehjati (commented):
Thanks for the reviews.

PR #142, opened 2022-10-14
Add set up details to development guidelines

MarahGamdou:
Added set up details to development guidelines to take into account possible set up issues

### DIFF
--- a/docs/development-guidelines.md
+++ b/docs/development-guidelines.md
@@ -6,7 +6,9 @@ You need to have:
 
 - [rootless Docker](https://docs.docker.com/engine/security/rootless/)
   - Note that if Docker is running as root on you local machine, you may run into some permission issues. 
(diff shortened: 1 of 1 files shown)


docs/development-guidelines.md
```diff
@@ -6,7 +6,9 @@ You need to have:
 
 - [rootless Docker](https://docs.docker.com/engine/security/rootless/)
   - Note that if Docker is running as root on you local machine, you may run into some permission issues. 
+  - Note that if you use Rootless mode, you might want to check that your Docker daemon is running properly, e.g., with `$ sudo systemctl status docker`. If your Docker daemon is inactive, you can start it manually using, e.g., with `$ sudo systemctl start docker`.
```
rbehjati, line 9: Is `sudo` necessary?

rbehjati, line 9: super-nit: The formatting script that we currently have does not automatically add line-breaks. Until we fix that, we have to add line-breaks manually. Can you add line breaks the new lines? :)

rbehjati (approved):
Nice. Thanks :)

docs/development-guidelines.md
MarahGamdou: Not for `$ sudo systemctl status docker`. I will remove it.

PR #151, opened 2022-11-09
Add FuzzClaim specification definition

MarahGamdou:
Add the definition of the claimSpec for the FuzzClaims

### DIFF
--- a/cmd/fuzzbinder/main.go
+++ b/cmd/fuzzbinder/main.go
@@ -1,0 +1,20 @@
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package main contains a command-line tool for generating fuzzing claims
+// for a revision of the source code.
+package main
+
+func main() {
+}
--- a/internal/fuzzbinder/fuzzbinder.go
+++ b/internal/fuzzbinder/fuzzbinder.go
@@ -14,5 +14,4 @@
 
 // Package fuzzbinder provides a function for generating a fuzzing claim
 // for a revision of the source code.
-
 package fuzzbinder
--- a/internal/fuzzbinder/fuzzbinder.go
+++ b/internal/fuzzbinder/fuzzbinder.go
@@ -1,0 +1,18 @@
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package fuzzbinder provides a function for generating a fuzzing claim
+// for a revision of the source code.
+
+package fuzzbinder
--- a/internal/fuzzbinder/fuzzclaim.go
+++ b/internal/fuzzbinder/fuzzclaim.go
@@ -1,0 +1,112 @@
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fuzzbinder
+
+// This file provides a custom `ClaimSpec` type, FuzzClaimSpec, to be used
+// for fuzzing claims within the ClaimPredicate (defined in amber package).
+// FuzzClaimSpec is intended to be used for providing the user with the
+// needed elements to characterize the security of a revision of the source
+// code based on fuzzing.
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/project-oak/transparent-release/pkg/amber"
+)
+
+// FuzzClaimV1 is the URI that should be used as the ClaimType in V1 Amber
+// Claim representing a V1 Fuzz Claim.
+const FuzzClaimV1 = "https://github.com/project-oak/transparent-release/fuzz_claim/v1"
+
+// FuzzClaimSpec gives the `ClaimSpec` definition.
+type FuzzClaimSpec struct {
+	// `ClaimSpec` per fuzz-target.
+	PerTarget []PerTargetSpec `json:"perTarget"`
+	// `ClaimSpec` for all the fuzz-targets.
+	PerProject *PerProjectSpec `json:"perProject"`
+}
+
+// PerTargetSpec contains the fuzzing claims specification per fuzz-target.
+type PerTargetSpec struct {
+	// Name of the fuzz-target.
+	Name string `json:"name"`
+	// URI of the fuzz-target.
+	URI string `json:"uri"`
+	// Coverage specifies the code coverage by the fuzz-target.
+	Coverage *FuzzCoverage `json:"coverage"`
+	// Bugs specifies the number of detected bugs using the fuzz-target.
+	Bugs int `json:"bugs"`
+	// Crashes specifies the number of crashes obtained using the fuzz-target.
+	Crashes int `json:"crashes"`
+	// FuzzEffort specifies the fuzzing efforts spent while using the fuzz-target.
+	FuzzEffort *FuzzEffortSpec `json:"fuzzEffort,omitempty"`
+}
+
+// PerProjectSpec contains the fuzzing claims specification of the revision
+// of the source code for all fuzz-targets.
+type PerProjectSpec struct {
+	// Coverage specifies the code coverage by all fuzz-targets.
+	Coverage *FuzzCoverage `json:"coverage"`
+	// Bugs specifies the  number of detected bugs using all fuzz-targets.
+	Bugs int `json:"bugs"`
+	// Crashes specifies the  number of crashes obtained using all fuzz-targets.
+	Crashes int `json:"crashes"`
+	// FuzzEffort specifies the fuzzing efforts spent while using all fuzz-targets.
+	FuzzEffort *FuzzEffortSpec `json:"fuzzEffort,omitempty"`
+}
+
+// FuzzCoverage contains the code coverage by fuzz testing.
+type FuzzCoverage struct {
+	// Line specifies line coverage.
+	Line string `json:"line"`
+	// Branch specifies branch coverage.
+	Branch string `json:"branch"`
+}
+
+// FuzzEffortSpec contains the fuzzing efforts.
+type FuzzEffortSpec struct {
+	// FuzzTime specifies the fuzzing time.
+	FuzzTime *time.Duration `json:"fuzzTime,omitempty"`
+	// NumberTests specifies the number of executed fuzzing tests.
+	NumberTests int `json:"numberTests,omitempty"`
+}
+
+// ValidateFuzzClaim validates that an Amber Claim is a Fuzz Claim with a valid ClaimType.
+// If valid, the ClaimPredicate object is returned. Otherwise an error is returned.
+func ValidateFuzzClaim(claimPredicate amber.ClaimPredicate) (*amber.ClaimPredicate, error) {
+	if claimPredicate.ClaimType != FuzzClaimV1 {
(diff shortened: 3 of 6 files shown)


internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,111 @@
...
+	// Crashes specifies the number of crashes obtained using the fuzz-target.
+	Crashes int `json:"crashes"`
+	// FuzzEffort specifies the fuzzing efforts spent while using the fuzz-target.
+	FuzzEffort *FuzzEffortSpec `json:"fuzzEffort,omitempty"`
+}
+
+// PerProjectSpec contains the fuzzing claims specification of the revision
+// of the source code for all fuzz-targets.
+type PerProjectSpec struct {
+	// Coverage specifies the code coverage by all fuzz-targets.
```
mariaschett, line 60: Do we want to include the name/uri of the project, too? Similar to the PerTargetSpec.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,111 @@
...
+	Coverage *FuzzCoverage `json:"coverage"`
```
mariaschett, line 61: IIUC there is duplication on `Coverage`, `Bugs`... between `PerTargetSpec` and `PerProjectSpec`. Would it make sense to extract these parts in a seperate type?

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,111 @@
...
+	default:
+		return nil, fmt.Errorf(
+			"the claimSpec does not have the expected type; got: %T, want: FuzzClaimSpec",
+			claimPredicate.ClaimSpec)
+	}
+}
+
+// validateFuzzClaimSpec validates details about the FuzzClaimSpec.
+func validateFuzzClaimSpec(claimPredicate amber.ClaimPredicate) (*amber.ClaimPredicate, error) {
+	// TBA
```
mariaschett, line 109: You could open an issue and add //TODO(#issue number): Short description.

mariaschett (approved):
Looks good to me! I am relying on @rbehjati for readability, though! :)

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// PerProjectSpec contains the fuzzing claims specification of the revision
+// of the source code for all fuzz-targets.
+type PerProjectSpec struct {
+	// Coverage specifies the code coverage by all fuzz-targets.
+	Coverage *FuzzCoverage `json:"coverage"`
+	// Bugs specifies the  number of detected bugs using all fuzz-targets.
+	Bugs int `json:"bugs"`
+	// Crashes specifies the  number of crashes obtained using all fuzz-targets.
```
rbehjati, line 64:
```suggestion
	// Crashes specifies the total number of crashes that happened when running the fuzz-targets.
```

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,111 @@
...
+	// Line specifies line coverage.
+	Line string `json:"line"`
+	// Branch specifies branch coverage.
+	Branch string `json:"branch"`
+}
+
+// FuzzEffortSpec contains the fuzzing efforts.
+type FuzzEffortSpec struct {
+	// FuzzTime specifies the fuzzing time.
+	FuzzTime *time.Duration `json:"fuzzTime,omitempty"`
```
rbehjati, line 81: I don't know how `time.Duration` is serialized into JSON. Does the serialization in JSON include the unit (seconds, minutes, etc)?

internal/fuzzbinder/fuzzbinder.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package fuzzbinder provides a function for generating a fuzzing claim
+// for a revision of the source code.
+package fuzzbinder
```
rbehjati, line 17: This file is not really needed, right? Or is the plan to add some functionality here in the future?

cmd/fuzzbinder/main.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package main contains a command-line tool for generating fuzzing claims
+// for a revision of the source code.
+package main
+
+func main() {
+}
```
rbehjati, line 20: I suggest adding something like `fmt.Println("Not yet implemented!")` to be clear that some functionality will be added here.

internal/fuzzbinder/fuzzclaim.go
rbehjati: This is a good point. I think `PerProjectSpec` is supposed to be an aggregation over `PerTargetSpec` for all the fuzz-targets. So this should be implicit, unless somehow there are fuzz-targets that are considered here, but not included in the list of `PerTargetSpec`s. 

I am not sure if that scenario is possible. In any case, I think it would be nice to state this relationship between `PerTarger` and `PreProject` fields. Is there a formula that could be used for deriving the `PerProject` from `PerTarget` list?

internal/fuzzbinder/fuzzbinder.go
MarahGamdou: Yes, I plan to add some functionality in the future there, but I will remove it for now and add it when it is needed. I will move the comment describing the package to the other file.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,111 @@
...
+	// `ClaimSpec` for all the fuzz-targets.
+	PerProject *PerProjectSpec `json:"perProject"`
+}
+
+// PerTargetSpec contains the fuzzing claims specification per fuzz-target.
+type PerTargetSpec struct {
+	// Name of the fuzz-target.
+	Name string `json:"name"`
+	// URI of the fuzz-target.
+	URI string `json:"uri"`
```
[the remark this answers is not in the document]
MarahGamdou: Do you mean that it's better to add "The URI must match the Git repository" in the comment?

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: Yes, `PerProject`  should include all the fuzz targets in the GitHub repo. It's not possible to derive the `PerProject`  from `PerTarget`  without using the evidence or fuzz reports because the aggregation of coverage depends on the intersection of covered lines/branches by the fuzz-targets.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I added `fmt.Println("Not yet implemented!")` to be clear that I will add it like for main.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: Thanks! I removed the duplication.

cmd/fuzzbinder/main.go
MarahGamdou: Using `fmt.Println` causes a linting error. It's included in forbidigo.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I replaced it with `int` (to use the time in seconds) for simplicity. However, it's possible to use the `time.Duration`  but not directly as described here: https://biscuit.ninja/posts/go-unmarshalling-json-into-time-duration/

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,106 @@
...
+	// Line specifies line coverage.
+	Line string `json:"line"`
+	// Branch specifies branch coverage.
+	Branch string `json:"branch"`
+}
+
+// FuzzEffortSpec contains the fuzzing efforts.
+type FuzzEffortSpec struct {
+	// FuzzTime specifies the fuzzing time in seconds.
+	FuzzTime int `json:"fuzzTime,omitempty"`
```
tiziano88, line 76: This is now ambiguous, as the field name does not specify the unit (it's in the comment, but it's better not to rely on comments for semantics). Consider `FuzzTimeSeconds` instead.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,106 @@
...
+type FuzzSpec struct {
+	// Name of the fuzz-target if FuzzSpec is used for one fuzz-target.
+	Name string `json:"name,omitempty"`
+	// Path of the fuzz-target, relative to the root of the Git repository,
+	// if FuzzSpec is used for one fuzz-target.
+	Path string `json:"path,omitempty"`
+	// Coverage specifies the code coverage by a fuzz-target or all fuzz-targets.
+	Coverage *FuzzCoverage `json:"coverage"`
+	// Bugs specifies the number of detected bugs using a fuzz-target or all fuzz-targets.
+	Bugs int `json:"bugs"`
```
tiziano88, line 56: Maybe `DetectedBugs` and `DetectedCrashes`? But up to you, the current is fine too.

tiziano88 (approved):

cmd/fuzzbinder/main.go
rbehjati: Sorry, you could use `log.Println` or `go.Printf`.

internal/fuzzbinder/fuzzclaim.go
rbehjati: I agree.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// FuzzSpec contains the fuzzing claims specification of a revision of
+// a source code per fuzz-target or for all fuzz-targets.
+type FuzzSpec struct {
+	// Name of the fuzz-target if FuzzSpec is used for one fuzz-target.
+	Name string `json:"name,omitempty"`
+	// Path of the fuzz-target, relative to the root of the Git repository,
+	// if FuzzSpec is used for one fuzz-target.
+	Path string `json:"path,omitempty"`
```
rbehjati, line 52: What would be the path in case of PreProject fuzz spec?

I think this is too much reuse. If it is only relevant for PerTarget, then this is not the best design in my opinion, and can cause confusion.

internal/fuzzbinder/fuzzclaim.go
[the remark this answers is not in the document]
MarahGamdou: I renamed it. thanks :))

internal/fuzzbinder/fuzzclaim.go
rbehjati: I prefer the previous design. You could alternatively have a separate struct with only the common parts, but then you have to come up with a name for that struct, perhaps `FuzzStats`.

So basically you'd have `FuzzSpecForTrarget`, `FuzzSpecForProject`, and `FuzzStats`. The first two use `FuzzStats`.

MarahGamdou: In this case,  `FuzzSpecForProject` is the same as `FuzzStats`, right? So,  two struct  `FuzzSpecForTrarget` and `FuzzStats` enough.  `FuzzSpecForTrarget`  uses `FuzzStats` and `FuzzStats` is used directly for `perProject`

rbehjati: Yes. That should work. Thanks :)

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I agree. Thanks!

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: Thanks! I used `DetectedBugs` instead of  `Bugs` . For the crashes, I'm not sure if we can use `Detected`  so I kept `Crashes`.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I added  `log.Println("Not yet implemented!")` to show that it is not yet implemented.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I modified the design based on your suggestion.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// FuzzSpecPerTarget contains the fuzzing claims specification per fuzz-target.
+type FuzzSpecPerTarget struct {
+	// Name of the fuzz-target.
+	Name string `json:"name"`
+	// Path of the fuzz-target, relative to the root of the Git repository.
+	Path string `json:"path"`
+	// Fuzzing statistics of the fuzz-target.
+	FuzzStatsPerTarget *FuzzStats `json:"fuzzStats"`
```
rbehjati, line 53: super nit: Feel free to ignore :)

```suggestion
	FuzzStats *FuzzStats `json:"fuzzStats"`
```

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// FuzzStats contains the fuzzing statistics of the revision
+// of the source code for all fuzz-targets or a fuzz-target.
+type FuzzStats struct {
```
rbehjati, line 58: Suggestion for future PRs: If flattening this and inlining `FuzzCoverage` and `FuzzEffortSpec` makes things easier in future PRs, feel free to consider making that change :)

rbehjati (approved):
Very nice. Thanks.

PR #153, opened 2022-11-22
Remove the use of bazel

### LINKED ISSUE #67: Remove bazel and convert the repo to a standard go project
We now have `go.mod` and `go.sum`, and can use most `go` commands, e.g., for building. Can we now remove all bazel-related files and convert this repo to a standard project? 

Currently when running `go test ./...` some `auth-logic` tests fail. I think this is because of the genrules that fetch and build the dependencies. Is it possible to replaces those bazel targets with go commands? (`go generate ./...` might be relevant).

WDYT? @aferr @tiziano88

rbehjati:
Fixes #67.


tiziano88 (approved):

MarahGamdou:
Nice changes

PR #157, opened 2022-11-25
Remove issuer from example

rbehjati:
Cleanup left over from #139.


MarahGamdou (approved):
Thanks for the changes!

PR #159, opened 2022-11-29
Add internal/intermediate struct for provenance data

mariaschett:
Adds and uses an internal intermediate struct `ProvenanceIR` in the verifier.

- [x] Tests pass
- [N/A] Appropriate changes to README are included in PR


internal/endorser/endorser.go
```diff
@@ -37,7 +38,9 @@ import (
 // valid. Each provenanceURI must either specify a local file (using the `file` scheme), or a
 // remote file (using the `http/https` scheme).
 func GenerateEndorsement(binaryDigest string, validityDuration amber.ClaimValidity, provenanceURIs []string) (*intoto.Statement, error) {
-	verifiedProvenances, err := loadAndVerifyProvenances(provenanceURIs, binaryDigest)
+	expected := verifier.ProvenanceIR{BinarySHA256Digest: []string{binaryDigest}}
```
rbehjati, line 41: How about instead passing `expected verifier.ProvenanceIR` as input to `GenerateEndorsement`? So that the caller of `GenerateEndorsement` is responsible for creating an instance of `verifier.ProvenanceIR`.

internal/endorser/endorser.go
```diff
@@ -93,9 +96,11 @@ func loadAndVerifyProvenances(provenanceURIs []string, binaryDigest string) (*am
 // binaryDigest as the subject. Provenances are consistent if they all have the same binary name and
 // binary digest. An error is returned if any of the conditions above are violated.
 // TODO(b/222440937): Document any additional checks.
-func verifyProvenances(provenances []slsa.ValidatedProvenance, binaryDigest string) error {
+func verifyProvenances(provenances []slsa.ValidatedProvenance, expected verifier.ProvenanceIR) error {
```
rbehjati, line 99:
```suggestion
func verifyProvenances(provenances []slsa.ValidatedProvenance, referenceValues verifier.ProvenanceIR) error {
```

Optional suggestion to be more specific :)

internal/verifier/verifier.go
```diff
@@ -78,10 +78,16 @@ func (verifier *ReproducibleProvenanceVerifier) Verify(provenanceFilePath string
 		return fmt.Errorf("couldn't build the binary: %v", err)
 	}
 
-	// The provenance is valid, therefore `expectedBinaryHash` is guaranteed to be non-empty.
-	expectedBinaryDigest := provenance.GetBinarySHA256Digest()
+	var actual ProvenanceIR
+	actual.FromAmber(provenance)
 
-	if err := buildConfig.VerifyBinarySHA256Digest(expectedBinaryDigest); err != nil {
+	expectedBinaryDigest, err := buildConfig.ComputeBinarySHA256Digest()
```
rbehjati, line 84: I don't quite see the benefit of this expansion. Can this logic be refactored to a `FromBuildConfig` function?

internal/verifier/verifier.go
```diff
@@ -122,3 +128,47 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify(provenanceFilePath strin
 
 	return nil
 }
+
+// Internal intermediate representation of Provenances for verification.
```
rbehjati, line 132: Golang comments start with the name of the element; in this case `ProvenanceIR`. The linters currently don't enforce it, but it is good to be consistent. Ref go/go-style/decisions#doc-comments.

internal/verifier/verifier.go
```diff
@@ -122,3 +128,47 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify(provenanceFilePath strin
...
+// We use the ProvenanceIR to
+// (1) map different provenance formats, and
+// (2) hold reference values.
+// To be usable with different provenance formats, we allow fields to be empty ([]) and to hold several reference values.
+type ProvenanceIR struct {
+	BinarySHA256Digest []string
+}
+
+func (p *ProvenanceIR) FromSLSA(provenance *slsa.ValidatedProvenance) {
```
rbehjati, line 141: I am not sure what golang conventions are, but I think if `FromSLSA` and `FromAmber` are functions, instead of methods on `ProvenanceIR` the code will be nicer. In particular, you'd have:

```
actual := verifier.FromSLSA(&provenances[index])
```

instead of 

```
var actual verifier.ProvenanceIR
actual.FromSLSA(&provenances[index])
```

rbehjati (commented):
Nice. Thanks :) 

My main comments are on the logic of `VerifyWithReference`. Let me know if you'd rather discuss separately.

internal/verifier/verifier.go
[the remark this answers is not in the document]
mariaschett: Hmmm... I have been thinking, but I don't see a strong reason for them. 

The mapping should be obvious from the code. 

We could use unit tests to document our assumptions, e.g., that "*amber.ValidatedProvenance contains a SHA256 hash of a single subject", but that should actually be implied by the validation.

Do you see a good case?

internal/verifier/verifier.go
```diff
@@ -122,3 +128,47 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify(provenanceFilePath strin
...
+
+func (p *ProvenanceIR) FromAmber(provenance *amber.ValidatedProvenance) {
+	// A *amber.ValidatedProvenance contains a SHA256 hash of a single subject.
+	p.BinarySHA256Digest = []string{provenance.GetBinarySHA256Digest()}
+}
+
+// VerifyWithReference verifies a provenance against a given reference.
+// We assume that we want to verify all fields the actual ProvenanceIR contains against the expected.
+// TODO(b/222440937): In future, also verify the details of the given provenance and the signature.
+func VerifyWithReference(actual ProvenanceIR, expected ProvenanceIR) error {
```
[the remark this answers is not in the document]
mariaschett: I'd like to use the `AmberProvenanceMetadataVerifier` in the next PR as wrapper for the `ProvenanceIRVerifier` in the endorser.

internal/verifier/verifier.go
```diff
@@ -122,3 +128,47 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify(provenanceFilePath strin
...
+	if expected.BinarySHA256Digest == nil {
+		return fmt.Errorf("no reference binary SHA256 digest given")
+	}
+
+	if actual.BinarySHA256Digest == nil {
+		// We don't want to verify the binary SHA256 digest.
```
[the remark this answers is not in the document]
mariaschett: I suggest the convention to only verify fields which are not empty in `Got`, i.e., all fields to be optional for flexibility in using the `ProvenanceIRVerifier`.

I agree that the digest should be present, but I think that should be caught by the `From` conversion/validation.

internal/verifier/verifier.go
```diff
@@ -122,3 +128,47 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify(provenanceFilePath strin
...
+		return nil
+	}
+
+	for _, expected := range expected.BinarySHA256Digest {
```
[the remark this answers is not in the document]
mariaschett: Added a comment, hope this clarifies :)

mariaschett (commented):
Thank you for comments.

internal/endorser/endorser.go
```diff
@@ -26,18 +26,19 @@ import (
...
 	"os"
 
+	"github.com/project-oak/transparent-release/internal/verifier"
 	"github.com/project-oak/transparent-release/pkg/amber"
 	"github.com/project-oak/transparent-release/pkg/intoto"
 	slsa "github.com/project-oak/transparent-release/pkg/intoto/slsa_provenance/v0.2"
 )
 
-// GenerateEndorsement generates an endorsement statement for the given binary SHA256 digest, for
+// GenerateEndorsement generates an endorsement statement for the given reference, for
```
rbehjati, line 35: hmm... this does not sound right now. We don't generate an Endorsement for a reference. 

```suggestion
// GenerateEndorsement generates an endorsement statement for the binary SHA256 digest in the given reference, for
```

internal/endorser/endorser.go
```diff
@@ -76,13 +77,13 @@ func loadAndVerifyProvenances(provenanceURIs []string, binaryDigest string) (*am
...
 
-	if err := verifyProvenances(provenances, binaryDigest); err != nil {
+	if err := verifyProvenances(provenances, reference); err != nil {
 		return nil, fmt.Errorf("verification of provenances failed: %v", err)
 	}
 
 	verifiedProvenances := amber.VerifiedProvenanceSet{
 		BinaryName:   provenances[0].GetBinaryName(),
-		BinaryDigest: binaryDigest,
+		BinaryDigest: reference.BinarySHA256Digests[0],
```
rbehjati, line 86: This does not sound right. If the reference can contain multiple values for `BinarySHA256Digests`, the value for `BinaryDigest` should not come from the reference, but from the provenances. We also need to make sure that if there are multiple provenances, they should all agree on the binary SHA256 digest.

internal/verifier/verifier.go
```diff
@@ -119,3 +119,65 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify() error {
...
+}
+
+// FromSLSAv0 maps data from a validated SLSAv0 provenance to ProvenanceIR.
+func FromSLSAv0(provenance *slsa.ValidatedProvenance) ProvenanceIR {
+	return ProvenanceIR{
+		// A slsa.ValidatedProvenance contains a SHA256 hash of a single subject.
+		BinarySHA256Digests: []string{provenance.GetBinarySHA256Digest()}}
+}
+
+// FromSLSAv0 maps data from a validated Amber provenance to ProvenanceIR.
```
rbehjati, line 139:
```suggestion
// FromAmber maps data from a validated Amber provenance to ProvenanceIR.
```

internal/verifier/verifier.go
```diff
@@ -119,3 +119,65 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify() error {
...
+
+// ProvenanceIRVerifier verifies a provenance against a given reference, by verifying
+// all non-empty fields in got using fields in want. Empty fields will not be verified.
+type ProvenanceIRVerifier struct {
+	Got  ProvenanceIR
+	Want ProvenanceIR
+}
+
+// TODO(b/222440937): In future, also verify the details of the given provenance and the signature.
+// We only verify fields which are not empty in Got, all empty fields are ignored.
```
rbehjati, line 154:
```suggestion
// Verify verifies an instance of ProvenanceIRVerifier by comparing its Got and Want fields. We only verify fields which are not empty in Got.
```

internal/verifier/verifier.go
```diff
@@ -119,3 +119,65 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify() error {
...
+// If a field in Got contains more than one value, we return an error.
+func (verifier *ProvenanceIRVerifier) Verify() error {
+	if len(verifier.Got.BinarySHA256Digests) != 1 {
+		return fmt.Errorf("got not exactly one actual binary SHA256 digest (%v)", verifier.Got.BinarySHA256Digests)
```
rbehjati, line 158:
```suggestion
		return fmt.Errorf("provenance must have exactly one binary SHA256 digest value, got (%v)", len(verifier.Got.BinarySHA256Digests))
```

internal/verifier/verifier.go
```diff
@@ -119,3 +119,65 @@ func (verifier *AmberProvenanceMetadataVerifier) Verify() error {
...
+	}
+	return verifier.Got.verifyBinarySHA256Digest(verifier.Want)
+}
+
+// verifyBinarySHA256Digest verifies that a single binary SHA256 is contained in the reference binary SHA256 digests.
```
rbehjati, line 163:
```suggestion
// verifyBinarySHA256Digest verifies that the binary SHA256 in this provenance is contained in the given reference binary SHA256 digests (in want).
```

internal/endorser/endorser_test.go
```diff
@@ -44,7 +45,10 @@ func TestGenerateEndorsement_SingleValidEndorsement(t *testing.T) {
 	}
 	tempURI := "file://" + tempPath
 	provenances := []string{tempURI}
-	statement, err := GenerateEndorsement(binaryHash, validity, provenances)
+	reference := verifier.ProvenanceIR{
```
rbehjati, line 48: Can you add a test with multiple values in `BinarySHA256Digests`?

rbehjati (commented):
Thanks for the fixes :) I think the logic for handling multiple reference values for the binary digest is still not quite right. Please see my comment on the changes in `loadAndVerifyProvenances`.

internal/verifier/verifier.go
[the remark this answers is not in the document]
rbehjati: It will make refactoring easier. When the logic gets more complicated, and if we decide to make changes to the mapping, having tests will make it easier. But we can also add tests once there are more fields in ProvenanceIR.

internal/endorser/endorser_test.go
mariaschett: Yes, I'll do that!

mariaschett (commented):
Thank you for the good review, Razieh! I'll add the tests on Tuesday.

rbehjati (approved):
Very nice. Thanks :)

internal/endorser/endorser.go
rbehjati: Yes. That makes sense. Thanks :)

rbehjati (approved):

internal/endorser/endorser_test.go
```diff
@@ -70,12 +74,15 @@ func TestGenerateEndorsement_MultipleValidEndorsement(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Could not load provenance: %v", err)
 	}
-	tempPath2, err := copyToTemp("../../testdata/provenance.json")
+	tempPath2, err := copyToTemp("../../testdata/provenance_variant.json")
```
mariaschett, line 77: @rbehjati I think in this test we just over-wrote the `provenance.json` in temp before. I created a variant of `provenance.json`. Does this make sense?

internal/endorser/endorser_test.go
```diff
@@ -92,13 +99,80 @@ func TestGenerateEndorsement_FailingSingleRemoteProvenanceEndorsement(t *testing
...
+		BinarySHA256Digests: []string{binaryHash},
+	}
+	_, err := GenerateEndorsement(reference, validity, provenances)
 	want := "could not load provenances"
 	if err == nil || !strings.Contains(err.Error(), want) {
 		t.Fatalf("got %q, want error message containing %q,", err, want)
 	}
 }
 
+func TestGenerateEndorsement_MultipleNotVerifiedOrConsistent(t *testing.T) {
```
mariaschett, line 112: Added tests. Could you have a look @rbehjati ?

mariaschett (commented):
Added tests.

internal/endorser/endorser_test.go
rbehjati: The previous approach was more flexible. Now we have an additional test file to maintain, and it does not add much value.  Is there any difference between `provenance.json` and `provenance_variant.json`?

internal/endorser/endorser_test.go
```diff
@@ -92,13 +99,80 @@ func TestGenerateEndorsement_FailingSingleRemoteProvenanceEndorsement(t *testing
...
+			provenancePaths: []string{"file://" + tempPath1, "file://" + tempPath2},
+			want:            "provenances are not consistent",
+		},
+		// Provenances do not contain reference binary SHA256 digest and are inconsistent.
+		{
+			reference:       []string{binaryHash + "_diff"},
+			provenancePaths: []string{"file://" + tempPath1, "file://" + tempPath2},
+			want:            "do not contain the actual binary SHA256 digest", // "provenances are not consistent"
+		},
+		// Provenances do not contain reference binary SHA256 digest and are inconsistent.
```
rbehjati, line 155: Remove this commented block of code?

internal/endorser/endorser_test.go
rbehjati: Thanks. This test checks several scenarios. Since the scenarios do not require heavy set up, I prefer separate test cases, one for each scenario. That way when a test fails it will be easier to figure out which scenario failed.

internal/endorser/endorser_test.go
```diff
@@ -92,13 +99,80 @@ func TestGenerateEndorsement_FailingSingleRemoteProvenanceEndorsement(t *testing
...
+		provenancePaths []string
+		want            string
+	}{
+		// Provenances do not contain reference binary SHA256 digest, but are consistent.
+		{
+			reference:       []string{binaryHash + "_diff"},
+			provenancePaths: []string{"file://" + tempPath1, "file://" + tempPath1Variant},
+			want:            "do not contain the actual binary SHA256 digest",
+		},
+		// Provenances contain reference binary SHA256 digest, but are inconsistent.
```
rbehjati, line 143: Don't we need a scenario where there are multiple consistent provenances, and multiple reference values? 
That is the most realistic scenario. 
The two interesting scenarios are (1) where provenances are consistent, and the digest is contained in the reference values, and (2) the provenances are consistent, but the digest is not contained in the reference values.

testdata/invalid_hash_provenance.json
```diff
@@ -1 +1,43 @@
-{"_type":"https://in-toto.io/Statement/v0.1","predicateType":"https://slsa.dev/provenance/v0.2","subject":[{"name":"test.txt","digest":{"sha256":"1234"}}],"predicate":{"builder":{"id":""},"buildType":"https://github.com/project-oak/transparent-release/schema/amber-slsa-buildtype/v1/provenance.json","invocation":{"configSource":{}},"buildConfig":{"command":["cp","testdata/static.txt","test.txt"],"outputPath":"test.txt"},"materials":[{"uri":"bash@sha256:9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9","digest":{"sha256":"9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9"}},{"uri":"https://github.com/project-oak/transparent-release","digest":{"sha1":"9b5f98310dbbad675834474fa68c37d880687cb9"}}]}}
\ No newline at end of file
+{
```
rbehjati, line 1: I think we needed provenances to be formatted according to canonical JSON, but if the code is not sensitive to it, it is probably not so important for now.

rbehjati (commented):
Thanks for adding the tests. I think there are some missing scenarios still.

internal/endorser/endorser_test.go
mariaschett: I think (1) we had already (`TestGenerateEndorsement_MultipleValidEndorsement`) and (2) is the first "// Provenances do not contain reference binary SHA256 digest, but are consistent."

rbehjati: (1) is different from `TestGenerateEndorsement_MultipleValidEndorsement` in that in `TestGenerateEndorsement_MultipleValidEndorsement` we have only a single reference value; I am looking for a scenario with multiple reference values for the digest.

(2) is more generic than `"// Provenances do not contain reference binary SHA256 digest, but are consistent."`, in that (2) has multiple reference values. But actually, I was a bit confused by the wording of that comment. The current wording sounds like that the "provenances do not contain a SHA256 digest", which is a different negative test.

internal/endorser/endorser_test.go
```diff
@@ -92,13 +99,80 @@ func TestGenerateEndorsement_FailingSingleRemoteProvenanceEndorsement(t *testing
...
+	tempPath2, err := copyToTemp("../../testdata/different_provenance.json")
+	if err != nil {
+		t.Fatalf("Could not load provenance: %v", err)
+	}
+	tests := []struct {
+		reference       []string
+		provenancePaths []string
+		want            string
+	}{
+		// Provenances do not contain reference binary SHA256 digest, but are consistent.
```
rbehjati, line 137:
```suggestion
		// Provenances do not contain the given reference binary SHA256 digest value, but are consistent.
```

internal/endorser/endorser_test.go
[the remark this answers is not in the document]
mariaschett: Yes, this is a confusing comment! Sorry, changed.

mariaschett:
I seperated the tests and I think I cover all interesting cases now:  `ConsistentNotVerified`, `InconsistentVerified`, and `InconsistentNotVerified`.

`ConsistentVerified` is covered by `MultipleValidEndorsement`. I don't think adding more reference values would add anything.

internal/endorser/endorser_test.go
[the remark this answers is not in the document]
rbehjati: oh! The `1234` is the invalid one... sorry about that.

rbehjati:
> I seperated the tests and I think I cover all interesting cases now: `ConsistentNotVerified`, `InconsistentVerified`, and `InconsistentNotVerified`.
> 

Many thanks. This is easier for me understand :) 

> `ConsistentVerified` is covered by `MultipleValidEndorsement`. I don't think adding more reference values would add anything.

I think it does because. There was a bug earlier where the first element in the reference values (`reference.BinarySHA256Digests[0]`) was used as the digest of the binary in the endorsement. That is the scenario that I wanted to be checked and prevented. I don't see how that is covered in the existing tests.

mariaschett:
> That is the scenario that I wanted to be checked and prevented.

Very true, thank you! I added multiple reference values in `MultipleValidEndorsement` ---and refactored to simplify tests.

internal/endorser/endorser_test.go
```diff
@@ -75,12 +72,17 @@ func TestGenerateEndorsement_MultipleValidEndorsement(t *testing.T) {
...
+		BinarySHA256Digests: []string{binaryHash + "_diff", binaryHash},
+	}
+	provenanceSet, err := loadAndVerifyProvenances(provenances, reference)
 	if err != nil {
 		t.Fatalf("Could not generate endorsement from %q: %v", provenances[0], err)
 	}
 
-	testutil.AssertEq(t, "binary hash", statement.Subject[0].Digest["sha256"], binaryHash)
+	testutil.AssertEq(t, "binary name", provenanceSet.BinaryName, binaryName)
+	testutil.AssertEq(t, "binary hash", provenanceSet.BinaryDigest, binaryHash)
```
rbehjati, line 85: Nice assertion :)

rbehjati (approved):
Perfect. Thanks. Nice refactoring :)

PR #161, opened 2022-12-01
Refactor ReproducibleProvenanceVerifier

mariaschett:
- Change interface of `Verifier` from `Verify(path)` to `Verify()`, which makes implementing a `Verifier` flexible.
- Make type of provenace of `ReproducibleProvenanceVerifier` explicit by adding the field `*amber.ValidatedProvenance`
- Seperate parsing the `*amber.ValidatedProvenance` from the Verifier.
- Fix naming of `verifer` flag from `config` to `provenance_path`

- [x] Tests pass

### DIFF
--- a/cmd/verifier/main.go
+++ b/cmd/verifier/main.go
@@ -24,16 +24,15 @@ import (
 )
 
 func main() {
-	// TODO(mschett): I think this is wrong.
-	buildConfigPathPtr := flag.String("config", "",
+	provenance_path := flag.String("provenance_path", "",
 		"Required - Path to SLSA provenance file of the Amber build type.")
 	gitRootDirPtr := flag.String("git_root_dir", "",
 		"Optional - Root of the Git repository. If not specified, sources are fetched from the repo specified in the config file.")
 	flag.Parse()
 
-	provenance, err := amber.ParseProvenanceFile(*buildConfigPathPtr)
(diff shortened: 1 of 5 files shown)


cmd/verifier/main.go
```diff
@@ -20,20 +20,28 @@ import (
 	"log"
 
 	"github.com/project-oak/transparent-release/internal/verifier"
+	"github.com/project-oak/transparent-release/pkg/amber"
 )
 
 func main() {
-	buildConfigPathPtr := flag.String("config", "",
+	provenancePath := flag.String("provenance_path", "",
```
rbehjati, line 27: Nice! Thanks.

internal/verifier/verifier.go
```diff
@@ -28,16 +28,15 @@ import (
 // ProvenanceVerifier defines an interface with a single method `Verify` for
 // verifying provenances.
 type ProvenanceVerifier interface {
-	// Verify verifies an Amber/SLSA provenance file in the given path.
```
rbehjati, line 31: Why is the comment removed? Is not it the convention to have a doc comment for these methods?

rbehjati (approved):
Nice. Thanks :)

internal/verifier/verifier.go
mariaschett: Fixed, thank you!

Issue #126, opened 2022-09-14
Refactor the Verify function for more flexible testing


rbehjati:
Really cool test! As idea: If we need the `bad_command_provenance.json` only in this test, shall we maybe hard code it as string in the test? Might also be faster to see the whole test then.

Similar with the `invalidHashProvenancePath`?

_Originally posted by @mariaschett in https://github.com/project-oak/transparent-release/pull/124#discussion_r970766129_

PR #164, opened 2022-12-12
Add build cmd not empty

mariaschett:
- Add a check that `build_cmd` is not empty.
- Refactor to use the `ProvenanceMetadataVerifier`.
- Extract `parseBuildConfig` and add `GetBuildCmd`.


internal/endorser/endorser.go
```diff
@@ -58,14 +57,14 @@ func loadAndVerifyProvenances(provenanceURIs []string, referenceValues verifier.
 	}
 
 	// load provenances from URIs
-	provenances := make([]slsa.ValidatedProvenance, 0, len(provenanceURIs))
+	provenances := make([]amber.ValidatedProvenance, 0, len(provenanceURIs))
```
rbehjati, line 60: I intentionally changed this from `amber` to `slsa` because we are deprecating `amber`. Please keep `slsa`.

internal/verifier/verifier.go
```diff
@@ -127,7 +127,8 @@ func chdir(dir string) {
 // AmberProvenanceMetadataVerifier verifies Amber provenances by comparing the
 // content of the provenance predicate against a given set of expected values.
 type AmberProvenanceMetadataVerifier struct {
-	provenanceFilePath string
+	Got  *amber.ValidatedProvenance
```
rbehjati, line 130: Please use `slsa.ValidatedProvenance` here.

mariaschett: SG, but `AmberProvenanceMetadataVerifier` might a bit of a confusing name then. Should I try to add a comment/issue explaining?

rbehjati: I agree. I think `ProvenanceMetadataVerifier` is a better name. We will no longer have an Amber provenance.

internal/endorser/endorser.go
rbehjati: In fact, `slsa.ValidatedProvenance` and `amber.ValidatedProvenance` are the same struct, but are located in two different packages. As discussed, it is better to have it (and the `ProvenanceIR`) in package `internal/common`. 

The reason I did not already put `ValidatedProvenance` in package `internal/common` was to avoid a circular dependency, and I thought we'd remove amber anyway. I can think of two ways to avoid the circular dependency:
- add a package common (or a similar name) in `pkg`, in addition to the one we currently have in `internal/commom`.
- Move more code from `amber` and `slsa` to `internal/common` to avoid the circular dependency. 

Please let me know if you want to discuss more or prefer me to do this refactoring :)

internal/endorser/endorser.go
```diff
@@ -58,14 +57,14 @@ func loadAndVerifyProvenances(provenanceURIs []string, referenceValues verifier.
...
 	provenancesData := make([]amber.ProvenanceData, 0, len(provenanceURIs))
 	for _, uri := range provenanceURIs {
 		provenanceBytes, err := getProvenanceBytes(uri)
 		if err != nil {
 			return nil, fmt.Errorf("couldn't load the provenance file from %s: %v", uri, err)
 		}
-		provenance, err := slsa.ParseProvenanceData(provenanceBytes)
+		provenance, err := amber.ParseProvenanceData(provenanceBytes)
```
rbehjati, line 67: I think it would be nicer to also replace this with a generic function that returns a `ProvenanceIR`. So we don't have an explicit dependency to `slsa`, `amber` or any other package. Would that work?

internal/endorser/endorser.go
rbehjati: Perhaps some of this generic provenance-related structure and functionality could go to `pkg/provenance`, and be made public as well.

internal/endorser/endorser.go
mariaschett: I think that would be a good idea. I would propose this in a different PR.

internal/endorser/endorser.go
mariaschett: Yes! :) I'll do this this in the next PR.

internal/verifier/verifier.go
```diff
@@ -124,36 +124,26 @@ func chdir(dir string) {
...
 }
 
-// Verify verifies a given Amber provenance file by checking its content
-// against the expected values specified in this
-// AmberProvenanceMetadataVerifier instance. Returns an error if any of the
-// values is not as expected. Otherwise returns nil, indicating success.
+// Verify verifies a given provenance file by checking its content against the expected values
+// ProvenanceMetadataVerifier instance.
 // TODO(#69): Check metadata against the expected values.
-// TODO(#126): Refactor and separate verification logic from the logic for reading the file.
```
rbehjati, line 139: Can #126 be closed now? Or are there more references to it?

internal/verifier/verifier.go
```diff
@@ -190,18 +190,31 @@ type ProvenanceIRVerifier struct {
...
+			return combinedResult, fmt.Errorf("provenance must have exactly one binary SHA256 digest value, got (%v)", verifier.Got.BinarySHA256Digests)
+		}
+		combinedResult.Combine(nextResult)
 	}
 
-	nextresult, err := verifier.Got.verifyBinarySHA256Digest(verifier.Want)
-	if err != nil {
-		return result, fmt.Errorf("provenance must have exactly one binary SHA256 digest value, got (%v)", verifier.Got.BinarySHA256Digests)
+	// Verify HasBuildCmd.
+	if verifier.Want.BuildCmds != nil {
```
rbehjati, line 208: When we load a toml file of reference values, which does not have a build command, is `BuildCmds` set to `nil` or empty array?

internal/verifier/verifier.go
```diff
@@ -232,3 +245,18 @@ func (got *ProvenanceIR) verifyBinarySHA256Digest(want ProvenanceIR) (Verificati
 
 	return result, nil
 }
+
+// verifyHasBuildCmd verifies that the build cmd is not empty.
+func (got *ProvenanceIR) verifyHasBuildCmd() (VerificationResult, error) {
```
rbehjati, line 250: Should this be renamed to `verifyHasOneBuildCmd` to be clear?

internal/verifier/verifier_test.go
```diff
@@ -93,6 +93,48 @@ func TestReproducibleProvenanceVerifier_badCommand(t *testing.T) {
 	}
 }
 
+func TestLoadAndVerifyProvenances_HasBuildCmd(t *testing.T) {
```
rbehjati, line 96: Perhaps add a test for `ProvenanceIRVerifier.Verify()` where `Want` contains an empty array for `BuildCmds`. I am not sure what we'd expect in that case.

rbehjati (approved):
Nice. Thanks.

internal/verifier/verifier.go
mariaschett: Yes, there are: The `ReproducibleProvenanceVerifier`.

internal/verifier/verifier_test.go
mariaschett: Again, my vision with #166 would be. If we have an empty array in `Want`, the `ProvenanceIRVerifier` does not verify that value.

mariaschett (commented):
Thank you!

internal/verifier/verifier_test.go
rbehjati: Will there be a way to verify that a command does not exist in the provenance?

mariaschett: What exactly would we want to check? 

I would have thought, whether a command exists depends on the format of the provenance, e.g., for SLSA v.02, `buildCmd` does not exist. So I am not sure what we can check.

The way I imagined we can check on whether the value should exists for a provenance format is through reference values (#166 ).

rbehjati: Actually I think a valid provenance (amber or slsav1) should always have a command. If the only thing we want to check about the build command is its presence, and not its exact value, then we probably don't even need a reference value for it.

The only thing we should consider in checking `hasBuildCommand` is the provenance version. In particular, in slsa02  we don't set the build command.

rbehjati: > I would have thought, whether a command exists depends on the format of the provenance, e.g., for SLSA v.02, buildCmd does not exist. So I am not sure what we can check.

True. So I don't think a reference value is very helpful in this case. Unless we want to check the exact value of the build command. 

> The way I imagined we can check on whether the value should exists for a provenance format is through reference values (https://github.com/project-oak/transparent-release/issues/166 ).

I think reference values are one source of information, but the provenance format/version should be taken into account as well. Is there going to be a reference value for the accepted provenance format/version (e.g., slsa02, slsa1)?

mariaschett: I added a draft PR to show how I thought the check on provenanceFormat/version (version is a better work) could work [#167](https://github.com/project-oak/transparent-release/pull/167).

NB: Sadly the changes are a bit hidden, because I need to base on this PR. I can submit this PR and we discuss on the next PR.

rbehjati: I think it is better to add a test for `ProvenanceIRVerifier.Verify()` where `Want` contains an empty array for `BuildCmds`, and merge this PR, and continue the discussion on #167.

mariaschett: Done.

Issue #165, opened 2022-12-12
Generalize `endorser` to work on `ProvenanceIR`


mariaschett:
Based on this [comment](https://github.com/project-oak/transparent-release/pull/164/files#r1046043774) we want the `endorser` to not depend on the provenance format--either SLSA or Amber---but on `ProvenanceIR`.

mariaschett:
Two questions, @rbehjati :

1. When verifying consistency of provenances, we rely on a `BinaryName`. Is a BinaryName always present, similar to the binary digest? Or only for some buildTypes?

2. When generalizing the endorser to work on `ProvenanceIR`: where do we want to make the decision on the buildType? I propose to either (a) change the interface of `GenerateEndorsement` to take an argument `toProvenanceIR func([]byte) (*common.ProvenanceIR, error)` similar to [here](https://github.com/project-oak/transparent-release/blob/4c7ee505df1d92f34f4069a7f401cc11530498dc/internal/endorser/endorser.go#L51). Alternatively, we could (b) give the `buildType` as argument to `GenerateEndorsement`. What do you think?

rbehjati:
> 1. When verifying consistency of provenances, we rely on a `BinaryName`. Is a BinaryName always present, similar to the binary digest? Or only for some buildTypes?

It is always present. But I don't think we need a reference value for it. If there are multiple input provenances, all must agree on the `BinaryName`. The `BinaryName` is copied to the endorsement. 

> 2. When generalizing the endorser to work on `ProvenanceIR`: where do we want to make the decision on the buildType? I propose to either (a) change the interface of `GenerateEndorsement` to take an argument `toProvenanceIR func([]byte) (*common.ProvenanceIR, error)` similar to [here](https://github.com/project-oak/transparent-release/blob/4c7ee505df1d92f34f4069a7f401cc11530498dc/internal/endorser/endorser.go#L51). Alternatively, we could (b) give the `buildType` as argument to `GenerateEndorsement`. What do you think?

I don't think any of this is needed. The `buildType` is already part of the provenance. So no additional parameter is needed for deciding which `buildType` to use.

At a high-level I can imaging a function `FromProvenance` similar to the following (in simplified Go!!):

```
func FromProvenance(provenance *ValidatedProvenance) (ProvenanceIR) {
  bt := provenance.BuildType() // all valid provenances must have a buildType
  switch(bt) {
      case AMBER: return FromAmber(provenance)
      case SLSA02: return FromSLSA02(provenance)
[...]

PR #167, opened 2022-12-13
Add validated provenance and reference values

### LINKED ISSUE #166: Add `ValidatedReferenceValues` and `ValidatedProvenanceValues`
The `ProvenanceIR` is essentially a dictionary between reference values and values from all possible provenance formats/versions.

However, depending on whether we are using the `ProvenanceIR` for reference values or values from the provenance format we have different invariants:
- For reference values we want at least one reference value for every value we want to verify to be present. The values we want to verify depend on the provenance format, which the caller needs to specify when validating reference values.
- For values from provenances we have at most one value for every value we want to verify. We allow 0 values, so we can map every possible provenance format to the `ProvenanceIR`.

I want to add two type aliases for `ProvenanceIR`: `ValidatedReferenceValues` and `ValidatedProvenanceValues` to signal these invariants.

Our `ProvenanceIRVerifier` currently verifies every _non-empty_ value in `ValidatedReferenceValue`.

If we impose the following conventions on `ValidatedReferenceValues`, we can use the same `ProvenanceIRVerifier` for every provenance format:

When validating the reference values, depending on the provenance format, we can additionally encode
- whether no value needs to be present (by setting the value to []). This allows to share the same reference values among different provenance formats and we can always add more reference values.
- whether any value at all should be present (by adding a default value).

Alternatively, we could  
- maintain a map from the struct fields of `ProvenanceIR` to bool to indicate whether they need to be verified for a given provenance format, but I don't see much benefit there,
- replace the generic `ProvenanceIRVerifier` by verifiers for every provenance format (e.g. `AmberProveanceVerifier`) calling the individual verification functions one by one, e.g., `verifyBinarySHA256Digest`, `verifyHasBuildCmd`. I think this has a lot of repetition.

mariaschett:
Fixes #166 .

### DIFF
--- a/internal/verifier/verifier.go
+++ b/internal/verifier/verifier.go
@@ -146,6 +146,38 @@ func (verifier *ProvenanceMetadataVerifier) Verify() (VerificationResult, error)
 	return provenanceVerifier.Verify()
 }
 
+// ValidatedProvenanceValues contain exactly one or no value in every field.
+type ValidatedProvenanceValues = ProvenanceIR
+
+// ValidatedReferenceValues contain at least 1 value for every field the provenanceFormat mandates to check,
+// and 0 values for every field that is to be ignored.
+type ValidatedReferenceValues = ProvenanceIR
+
+// If a new field is added to `ProvenanceIR`, this has to be updated.
+func (want *ProvenanceIR) ValidateReferenceValues(provenanceFormat string) (*ValidatedReferenceValues, error) {
+
+	// BinarySHA256Digest
+	switch provenanceFormat {
+	default:
+		if len(want.BinarySHA256Digests) <= 1 {
+			return want, fmt.Errorf("no reference binary SHA256 digests given")
+		}
+	}
+
+	// BuildCmds
+	switch provenanceFormat {
+	case "Amber":
+	case "SLSAv1":
+		// By adding a default value the build cmd is checked by Verify, even if no build cmd is given in the reference values.
+		want.BuildCmds = [][]string{{"_"}}
+	case "SLSAv0":
+		// Even if the build cmd is given in the reference values, we want to ignore them.
+		want.BuildCmds = [][]string{{}}
+	}
+
+	return want, nil
+}
+
 // ProvenanceIR is an internal intermediate representation of data from provenances for verification.
 // We use the ProvenanceIR to
 // (1) map different provenance formats, and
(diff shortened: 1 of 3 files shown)


internal/verifier/verifier.go
```diff
@@ -146,6 +146,38 @@ func (verifier *ProvenanceMetadataVerifier) Verify() (VerificationResult, error)
...
+// and 0 values for every field that is to be ignored.
+type ValidatedReferenceValues = ProvenanceIR
+
+// If a new field is added to `ProvenanceIR`, this has to be updated.
+func (want *ProvenanceIR) ValidateReferenceValues(provenanceFormat string) (*ValidatedReferenceValues, error) {
+
+	// BinarySHA256Digest
+	switch provenanceFormat {
+	default:
+		if len(want.BinarySHA256Digests) <= 1 {
```
rbehjati, line 162: The condition should be `len(want.BinarySHA256Digests) < 1`, right?

internal/verifier/verifier.go
```diff
@@ -146,6 +146,38 @@ func (verifier *ProvenanceMetadataVerifier) Verify() (VerificationResult, error)
...
 
+// ValidatedProvenanceValues contain exactly one or no value in every field.
+type ValidatedProvenanceValues = ProvenanceIR
+
+// ValidatedReferenceValues contain at least 1 value for every field the provenanceFormat mandates to check,
+// and 0 values for every field that is to be ignored.
+type ValidatedReferenceValues = ProvenanceIR
+
+// If a new field is added to `ProvenanceIR`, this has to be updated.
+func (want *ProvenanceIR) ValidateReferenceValues(provenanceFormat string) (*ValidatedReferenceValues, error) {
```
rbehjati, line 157: Where does `provenanceFormat` come from? 
All provenance formats are slsa provenances, with different buildTypes. I suggest we use the buildType as the format, and have a reference value for the supported buildTypes as well. WDYT @mariaschett?

mariaschett: > Where does provenanceFormat come from?

Options: 
1. We infer the provenanceFormat from the given provenance.
2. The caller decides, which implicitly happens currently, when calling `ProvenanceMetadataVerifier` (we now that this is `SLSAv02`).

> buildType as format / reference value for the buildTypes

Let's discuss this offline.

mariaschett:
This is redudant because of #194 and #165 .

PR #168, opened 2022-12-13
Add FuzzBinder  cmd line tool

MarahGamdou:
- Add FuzzBinder cmd line tool to generate fuzzing claims from OSS-Fuzz and ClusterFuzz reports.
- Add how-to use FuzzBinder.
- Update fuzzing claim example in the testdata


internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+func generateFuzzClaimSpec(date string, revisionDigest intoto.DigestSet, fuzzParameters *FuzzParameters, fuzzTargets []string) (*FuzzClaimSpec, error) {
+	var fuzzClaimSpec FuzzClaimSpec
+	var projectCrashes Crash
+	var projectFuzzEffort FuzzEffort
+	fuzzersCrashes := make(map[string]*Crash)
+	fuzzersFuzzEffort := make(map[string]*FuzzEffort)
+	fuzzersCoverage := make(map[string]*Coverage)
+	hyphenDate := fmt.Sprintf("%s-%s-%s", date[:4], date[4:6], date[6:])
+	//Get fuzzing statistics.
+	projectCoverage, err := GetCoverage(date, fuzzParameters, "perProject", "")
```
rbehjati, line 196: It is better to refactor the code so that you don't have to pass a magic string like "perProject" to `GetCoverage`.

In particular, you could separate between `GetTotalCoverage` and `GetCoverageForTarget`. Whatever these two functions have in common should be factored into separate function they both call.

cmd/fuzzbinder/main.go
```diff
@@ -17,9 +17,54 @@
...
+	"path/filepath"
+
+	"github.com/project-oak/transparent-release/internal/fuzzbinder"
 )
 
 func main() {
-	log.Println("Not yet implemented!")
+	fuzzParameters := &fuzzbinder.FuzzParameters{}
+	var fuzzClaimPath string
+	var date string
```
rbehjati, line 32: Why is not date one of the `FuzzParameters`? In fuzzclaim.go it is often used together with FuzzParameters.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, fuzzTarget := range fuzzTargets {
+		coverage, err := GetCoverage(date, fuzzParameters, "perTarget", fuzzTarget)
+		if err != nil {
+			return nil, err
+		}
+		fuzzEffort, err := GetFuzzEffort(revisionDigest, hyphenDate, fuzzTarget, fuzzParameters)
```
rbehjati, line 206: Readability: It is better to use a consistent ordering among parameters everywhere. For instance, always first pass date, then the revision, then the parameters...

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
 
 	return &statement, nil
 }
+
+// generateFuzzClaimSpec generates a fuzzing claim specification using the
+// fuzzing reports of OSS-Fuzz.
+func generateFuzzClaimSpec(date string, revisionDigest intoto.DigestSet, fuzzParameters *FuzzParameters, fuzzTargets []string) (*FuzzClaimSpec, error) {
```
rbehjati, line 187: This function is too long and does too many things. Can you split into smaller functions?
There is a long distance between variable declaration and where they are used (e.g., in case of projectCoverage). 
I suggest you move the two for-loops for perTarget coverage into their own functions, and call them here for better readability.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+
+		projectCrashes.Detected = projectCrashes.Detected || crash.Detected
+		projectFuzzEffort.FuzzTimeSeconds += fuzzEffort.FuzzTimeSeconds
+		projectFuzzEffort.NumberFuzzTests += fuzzEffort.NumberFuzzTests
+	}
+	// Generate fuzzing claim specification.
+	fuzzClaimSpec.PerProject = &FuzzStats{
+		BranchCoverage:  projectCoverage.BranchCoverage,
+		LineCoverage:    projectCoverage.LineCoverage,
+		DetectedCrashes: projectCrashes.Detected,
```
rbehjati, line 227: Are you computing this and the following two values based on perTarget results because they don't exist perProject, or because it is more accurate this way?

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+			Name:      fmt.Sprintf("%s_%s_%s", fuzzParameters.FuzzEngine, fuzzParameters.ProjectName, fuzzTagret),
+			Path:      fmt.Sprintf("%s/fuzz/fuzz_targets/%s.rs", fuzzParameters.ProjectName, fuzzTagret),
+			FuzzStats: &targetStats,
+		}
+		fuzzClaimSpec.PerTarget = append(fuzzClaimSpec.PerTarget, targetSpec)
+	}
+	return &fuzzClaimSpec, nil
+}
+
+// GenerateFuzzClaim generates a fuzzing claim (an instance of intoto.Statement,
```
rbehjati, line 249: Nice comment :)

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+// with AmberClaimV1 as the PredicateType and FuzzClaimV1 as the ClaimType) using the
+// fuzzing reports of OSS-Fuzz.
+func GenerateFuzzClaim(date string, fuzzParameters *FuzzParameters) (*intoto.Statement, error) {
+	var statement intoto.Statement
+	var predicate amber.ClaimPredicate
+	revisionDigest, err := GetCoverageRevision(date, fuzzParameters)
+	if err != nil {
+		return nil, err
```
rbehjati, line 257: Can you double check that `err` contains all the required info, and if needed add some context to the error message? (see go/go-style/best-practices#error-extra-info)

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+	}
+	fuzzTargets, err := GetFuzzTargets(date, fuzzParameters)
+	if err != nil {
+		return nil, err
```
rbehjati, line 261: Same as above :)

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+	}
+	// Generate Amber predicate
+	predicate.ClaimType = FuzzClaimV1
+	currentTime := time.Now()
+	tomorrow := time.Now().AddDate(0, 0, 1)
+	nextWeek := time.Now().AddDate(0, 0, 7)
```
rbehjati, line 267: I guess we don't want these claims to be valid for only 7 days. I think the validity duration should be an input parameter. 

Feel free to add a TODO here and fix it in a future PR.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+	}
+	return &fuzzClaimSpec, nil
+}
+
+// GenerateFuzzClaim generates a fuzzing claim (an instance of intoto.Statement,
+// with AmberClaimV1 as the PredicateType and FuzzClaimV1 as the ClaimType) using the
+// fuzzing reports of OSS-Fuzz.
+func GenerateFuzzClaim(date string, fuzzParameters *FuzzParameters) (*intoto.Statement, error) {
+	var statement intoto.Statement
+	var predicate amber.ClaimPredicate
```
rbehjati, line 254: I think it is more conventional to create and initialize these variable at the end of the function, instead of having these declarations here.

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+		return nil, err
+	}
+	predicate.Evidence = evidences
+	// Generate intoto statement
+	statement.Type = intoto.StatementInTotoV01
+	subject := intoto.Subject{
+		Name:   fuzzParameters.ProjectHomepage,
+		Digest: revisionDigest,
+	}
+	statement.Subject = append(statement.Subject, subject)
```
rbehjati, line 289: This is strange. statement.Subject is empty, right? 

initializing statement instead of a separate declaration at the top, and gradual assignment of values is more readable. Something like the following:

```
statementHeader := intoto.StatementHeader{
  Type:          intoto.StatementInTotoV01,
  PredicateType: ...
  Subject:       []intoto.Subject{subject},
}

statement := intoto.Statement{
	StatementHeader: statementHeader,
	Predicate:       predicate,
}
```

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -180,3 +181,122 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
+	statement.PredicateType = amber.AmberClaimV1
+	validFuzzPredicate, err := ValidateFuzzClaim(predicate)
+	if err != nil {
+		return nil, err
+	}
+	statement.Predicate = *validFuzzPredicate
+	validAmberPredicate, err := amber.ValidateAmberClaim(statement)
```
rbehjati, line 296: This is strange. Are `ValidateFuzzClaim` and `ValidateAmberClaim` really needed? I think `ValidateFuzzClaim` subsumes `ValidateAmberClaim`. If they are both actually needed could you please add a comment explaining why?

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -102,9 +100,10 @@ type Crash struct {
 // FuzzParameters contains the fuzzing parameters
 // used in OSS-Fuzz project config.
 type FuzzParameters struct {
-	ProjectName string
-	FuzzEngine  string
-	Sanitizer   string
+	ProjectName     string
+	ProjectHomepage string
```
rbehjati, line 104: What is this? Can you add some documentation about it?

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -366,28 +381,48 @@ func GetFuzzTargets(fuzzParameters *FuzzParameters, date string) ([]string, erro
 
 // GetEvidences gets the list of the evidence files used by the fuzzscraper.
 // The expected date format is "YYYYMMDD".
-func GetEvidences(date string, fuzztargets []string, fuzzParameters *FuzzParameters) map[string]string {
-	evidences := make(map[string]string)
+func GetEvidences(date string, fuzzTargets []string, fuzzParameters *FuzzParameters) ([]amber.ClaimEvidence, error) {
+	evidences := make([]amber.ClaimEvidence, len(fuzzTargets)+2)
```
rbehjati, line 385: Why is a fixed-size array declared here? What is the benefit of it?

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -366,28 +381,48 @@ func GetFuzzTargets(fuzzParameters *FuzzParameters, date string) ([]string, erro
...
+	blobName = fmt.Sprintf("%s/reports/%s/linux/summary.json", fuzzParameters.ProjectName, date)
+	uri = fmt.Sprintf("gs://%s/%s", CoverageBucket, blobName)
+	digest, err = getGCSFileDigest(CoverageBucket, blobName)
+	if err != nil {
+		return nil, err
+	}
+	evidences[1] = amber.ClaimEvidence{Role: role, URI: uri, Digest: *digest}
+	for idx, fuzzTarget := range fuzzTargets {
+		// The role of the coverage evidence using the fuzzTarget.
+		role := fmt.Sprintf("%s_%s_%v coverage", fuzzParameters.FuzzEngine, fuzzParameters.ProjectName, fuzzTarget)
```
rbehjati, line 408: Do the roles have to be different?

Also, why is it needed to include the project name in the role? Is not that the same for all targets?
I think all these could have the same role: "Fuzz target". What do you think?

cmd/fuzzbinder/README.md
```diff
@@ -1 +1,64 @@
 # Fuzzing Claims for revisions of a source code
+
+The goal of Fuzzbinder is to generated, sign and publish in a transparency log fuzzing claims for revisions of the source code of a given project to endorse their security.
```
rbehjati, line 3:
```suggestion
The goal of Fuzzbinder is to generate, sign and publish in a transparency log fuzzing claims for revisions of the source code of a given project.
```

The endorsement is a separate step :)

cmd/fuzzbinder/README.md
```diff
@@ -1 +1,64 @@
...
+
+```sh
+$ gcloud auth login
+```
+
+If you still have authentication problems look at this [authentication guide](https://googleapis.dev/python/google-api-core/latest/auth.html) and try again. Once `gsutil` works, you can run the fuzzing claim generation command.
+
+### 2) Generate fuzzing claim
+
+The [`fuzzbinder`](/internal/fuzzbinder/) package provides functionality for generating a fuzzing claim file.
```
rbehjati, line 42: Is this path correct? I am not sure how absolute paths work on GitHub.

rbehjati: Also, `internal/fuzzbinder` is internal implementation details. The user should not need to know about this. If you want to keep it, I suggest you move it to the end of the document, as a note for further info.

cmd/fuzzbinder/README.md
```diff
@@ -1 +1,64 @@
...
+
+To generate a fuzzing claim run:
+
+```sh
+$ go run cmd/fuzzbinder/main.go -project_name <project-name> \
+   -homepage <path-to-project-git-repo-root>  \
```
rbehjati, line 48: Rename homepage to `git-repo`. Also elsewhere :)

cmd/fuzzbinder/README.md
```diff
@@ -1 +1,64 @@
...
+  -fuzzengine <fuzzing-engine> -sanitizer <fuzzing-sanitizer>  \
+  -date <fuzzing-date> -fuzzclaim_path <fuzzclaim-path>
+```
+
+As you may noticed, you will need `<fuzzing-engine>` and `<fuzzing-sanitizer>` that you used in your OSS-Fuzz project configuration.
```
rbehjati, line 53:
```suggestion
As you may have noticed, you will need `<fuzzing-engine>` and `<fuzzing-sanitizer>` that you used in your OSS-Fuzz project configuration.
```

cmd/fuzzbinder/README.md
```diff
@@ -1 +1,64 @@
...
+For instance, if you used `libfuzzer` as a fuzzing engine and an `address sanitizer` for your project, and you want to generate a fuzzing claim
```
rbehjati, line 54: Some of this info should be included in the help string of the command flags. Please see my comments on that.

rbehjati (commented):
Thanks Marah. Please see my comments, and let me know if you want to have a chat.

cmd/fuzzbinder/README.md
MarahGamdou: The path is correct. However, I removed it since it is not relevant for the how-to document as you mentioned.

cmd/fuzzbinder/README.md
MarahGamdou: done

cmd/fuzzbinder/main.go
MarahGamdou: Date is not in `FuzzParameters` because the expected date format is not always the same for all functions. In fact, OSS-Fuzz and ClusterFuzz expect different date formats.

However, I added it and made the expected date format the same for all functions for simplicity.

cmd/fuzzbinder/main.go
```diff
@@ -17,9 +17,54 @@
...
+	flag.StringVar(&fuzzParameters.ProjectName, "project_name", "",
+		"Required - Project name as defined in OSS-Fuzz projects.")
+	flag.StringVar(&fuzzParameters.ProjectHomepage, "homepage", "",
+		"Required - Project homepage.")
```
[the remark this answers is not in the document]
MarahGamdou: It is the GitHub repository of the project. I modified it.

cmd/fuzzbinder/main.go
```diff
@@ -17,9 +17,54 @@
...
+	flag.StringVar(&fuzzParameters.FuzzEngine, "fuzzengine", "",
+		"Required - Fuzzing engine.")
```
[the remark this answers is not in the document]
MarahGamdou: I added the possibilities of the fuzzing engine. Thanks!

cmd/fuzzbinder/main.go
```diff
@@ -17,9 +17,54 @@
...
+	flag.StringVar(&fuzzParameters.Sanitizer, "sanitizer", "",
+		"Required - Fuzzing sanitizer.")
```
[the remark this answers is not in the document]
MarahGamdou: I specified the possible values and a default value as "asan". Since it is one of the default sanitizers used by OSS-Fuzz.

cmd/fuzzbinder/main.go
```diff
@@ -17,9 +17,54 @@
...
+	flag.StringVar(&date, "date", "",
+		"Required - Fuzzing date.")
```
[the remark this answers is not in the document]
MarahGamdou: Done!

cmd/fuzzbinder/main.go
```diff
@@ -17,9 +17,54 @@
...
+	flag.StringVar(&fuzzClaimPath, "fuzzclaim_path", "",
+		"Required - Output file path for storing the generated fuzzing claim.")
+	flag.Parse()
+
+	// Get the absolute path for storing the fuzzing the claim.
```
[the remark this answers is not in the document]
MarahGamdou: Thanks!

mariaschett (approved):
LGTM!

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I added a `// TODO(#171): Split GetCoverage into GetTotalCoverage and GetCoverageForTarget.`

MarahGamdou: I added a related issue to rename functions that take a log of computing: `TODO(#172): Rename functions that take a lot of computation.`

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I added a related issue to split  this function `// TODO(#171): Split generateFuzzClaimSpec into smaller functions.`

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I added a TODO: `TODO(#173): Add validity duration as an input parameter.`

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: I added a documentation for all of them and replaced the `ProjectHomepage` by the `ProjectGitRepo`.

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -366,28 +381,48 @@ func GetFuzzTargets(fuzzParameters *FuzzParameters, date string) ([]string, erro
...
+	}
+	evidences[0] = amber.ClaimEvidence{Role: role, URI: uri, Digest: *digest}
+
 	// Get the GCS absolute path of the file containing the coverage summary for the project on a given day.
-	evidences["project coverage"] = fmt.Sprintf("gs://%s/%s/reports/%s/linux/summary.json", CoverageBucket, fuzzParameters.ProjectName, date)
-	for _, fuzztarget := range fuzztargets {
-		// The role of the coverage evidence using the fuzztarget.
-		role := fmt.Sprintf("%s_%s_%v coverage", fuzzParameters.FuzzEngine, fuzzParameters.ProjectName, fuzztarget)
+	role = "project coverage"
+	blobName = fmt.Sprintf("%s/reports/%s/linux/summary.json", fuzzParameters.ProjectName, date)
```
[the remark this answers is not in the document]
MarahGamdou: I added an issue and a TODO `TODO(#174): Replace GCS path by Ent path in evidences URI.`

cmd/fuzzbinder/README.md
```diff
@@ -1 +1,66 @@
...
+```
+
+Note that the coverage reports in `gs://oss-fuzz-coverage` are public. However, you need to be a primary_contact or an auto_cc for the project for which you want to generate fuzzing claims to have access to the fuzzing logs in `gs://${PROJECT_NAME}-logs.clusterfuzz-external.appspot.com` of that project. If you are not sure whether you are a primary_contact or an auto_cc for the project, see this [OSS-Fuzz documentation](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary) for more details.
+
+If you are a primary_contact or an auto_cc and you get an authorization error from the commands above, run the following commands:
+
+```sh
+$ gcloud auth application-default login
+```
+
```
rbehjati, line 34: It he next step required too?

Also, to complete the command above you have to follow a link to a browser and copy the token. Is not there a better way to authenticate?

cmd/fuzzbinder/README.md
```diff
@@ -1 +1,66 @@
...
+The goal of Fuzzbinder is to generate, sign and publish (in a transparency log) fuzzing claims for revisions of the source code of a given project.
+
+By following the steps described below, you will be able to generate fuzzing claims for your open-source project using the fuzzing reports generated by OSS-Fuzz and ClusterFuzz for your project, and the command line tool
+[`cmd/fuzzbinder`](/cmd/fuzzbinder/) that we have provided to generate fuzzing claims.
+
+## Generating fuzzing claims using the `cmd/fuzzbinder` tool
+
+If you have not already done so, consult the [`development guidelines`](./../../docs/development-guidelines.md) to find out what you need to install before following the steps below.
+
+### Step 1: Establish access to Google Cloud Storage
```
rbehjati, line 12: I suggest you split this into 3 steps:

Step 1: Get access to ClusterFuzz crash reports and fuzzer statistics 
See https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary

Step 2 (optional): install google cloud sdk

Step 3: Authenticate to google cloud

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: Done :) I used revisionDigest, fuzzParameters, then other parameters...

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: The perTarget results are also computed using all fuzzers logs (they are not extracted direclty). In the same way, we can't extract the perProject directly from the logs. 

OSS-Fuzz is also computing the fuzzing efforts and showing them in the html reports, but they are aggregated for all the day (even if multiple revisions are used). So, they can be used.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I agree. I moved them :)

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I agree :) I updated it.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: I added `ValidateAmberClaim` to `ValidateFuzzClaim` so that we don't have to call both. 

They are not really necessary since the generated claim should be valid by design, but it helps to check if there are problems (so it's just there as an extra precaution). I can remove them though.

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: I just added it to avoid linting error `Consider pre-allocating ...` but I modified it slightly to make it nicer.

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: The roles here are related to the extracted data. For example, some evidence is for project coverage and some for fuzzer coverage. So I tried to give them different roles depending on their use. My goal here is to help reproduce the code since the files can be very similar.

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -366,28 +381,48 @@ func GetFuzzTargets(fuzzParameters *FuzzParameters, date string) ([]string, erro
...
 // GetEvidences gets the list of the evidence files used by the fuzzscraper.
 // The expected date format is "YYYYMMDD".
-func GetEvidences(date string, fuzztargets []string, fuzzParameters *FuzzParameters) map[string]string {
-	evidences := make(map[string]string)
+func GetEvidences(date string, fuzzTargets []string, fuzzParameters *FuzzParameters) ([]amber.ClaimEvidence, error) {
+	evidences := make([]amber.ClaimEvidence, len(fuzzTargets)+2)
 	// Get the GCS absolute path of the file containing the revision hash of the source code used
 	// in the coverage build on a given day.
-	evidences["revision"] = fmt.Sprintf("gs://%s/%s/srcmap/%s.json", CoverageBucket, fuzzParameters.ProjectName, date)
+	role := "revision"
```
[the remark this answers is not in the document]
rbehjati: Also, I am not sure what the role "revision" mean.

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: > I think all these could have the same role: "Fuzz target". What do you think?
Each fuzz target has its one coverage summary.

cmd/fuzzbinder/main.go
```diff
@@ -13,13 +13,60 @@
...
+	// TODO(#175): Remove fuzzEngine and sanitizer from FuzzBinder inputs.
+	flag.StringVar(&fuzzParameters.FuzzEngine, "fuzzengine", "libFuzzer",
+		"Required - Fuzzing engine used for the project. Examples: libFuzzer, afl, honggfuzz, centipede.")
+	flag.StringVar(&fuzzParameters.Sanitizer, "sanitizer", "asan",
+		"Required - Fuzzing sanitizer used for the project. Examples: asan, ubsan, msan.")
+	flag.StringVar(&fuzzParameters.Date, "date", "",
+		"Required - Fuzzing date. The expected date format is YYYYMMDD. \n"+
+			"This date must not be more than 15 days prior to the date of execution of this cmd. \n"+
+			"Therefore, if you run this cmd on 20221215, the date must be between 20221215 and 20221201 (both dates included).")
+	var fuzzClaimPath string
```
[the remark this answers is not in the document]
MarahGamdou: Thanks !

internal/fuzzbinder/fuzzclaim.go
rbehjati: Looking at the implementation of `GetCoverageRevision`, adding context that this is about getting coverage revision is not repetitive.

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -263,12 +263,13 @@ func getFuzzEffortFromFile(reader io.Reader, revisionDigest intoto.DigestSet) (*
 		if err := lineScanner.Err(); err != nil {
 			return nil, err
 		}
-		fileFuzzEffort, err = getFuzzStatsFromFile(lineScanner)
+		gotFuzzEffort, err := getFuzzStatsFromFile(lineScanner)
```
rbehjati, line 267: Here you could just return 

```
return getFuzzStatsFromFile(lineScanner)
```

The function will be a lot simpler, and no separate variable declaration will be needed.

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: Done !

internal/fuzzbinder/fuzzclaim.go
MarahGamdou: Done !

internal/fuzzbinder/fuzzscraper.go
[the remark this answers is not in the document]
MarahGamdou: > One important evidence that we want to have is the source code for the fuzz target. Right? We don't have that. Can you add a todo for adding that?
It is supposed to be easy to find using the commit hash since it has to be in the project itself, right?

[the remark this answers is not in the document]
MarahGamdou: > Also, I am not sure what the role "revision" mean.

The evidence that helped to link the data to a given revision of the source code.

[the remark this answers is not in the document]
rbehjati: > It is supposed to be easy to find using the commit hash since it has to be in the project itself, right? Do we still need to have it as evidence if it is on GitHub?

You are right. We have the exact path relative to the repo root in the perTarget reports, right?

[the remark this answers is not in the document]
rbehjati: > The evidence that helped to link the data to a given revision of the source code. So, the evidence that helps get the commit hash.

Can you clearly document it somewhere? Also perhaps give it a better name :) What is the name of the file or bucket where this data came from?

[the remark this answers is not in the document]
MarahGamdou: > You are right. We have the exact path relative to the repo root in the perTarget reports, right?

Yes, at least for oak. I'm not sure if the other projects using OSS-Fuzz have to put the fuzz-targets in `fuzz/fuzz_targets/`. I can add it as an input if that's not the case.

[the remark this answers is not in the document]
rbehjati: How do get these paths? Cannot they be retrieved from the logs or other reports? I'd rather not add another input.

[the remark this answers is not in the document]
MarahGamdou: > Can you clearly document it somewhere? Also perhaps give it a better name :) What is the name of the file or bucket where this data came from?

I will document this in the design doc. It is in `oss-fuzz-coverage` is a `srcmap` folder. I can name it `map date-revision`. What do you think?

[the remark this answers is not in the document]
rbehjati: Somehow I feel like `srcmap` is better, but up to you. I think a clear documentation is more important.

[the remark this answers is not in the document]
MarahGamdou: > How do get these paths? Cannot they be retrieved from the logs or other reports? I'd rather not add another input.

I'm not sure if they are available somewhere in the GCS bucket (I didn't see them). They are passed to OSS-Fuzz using `--target-dir`.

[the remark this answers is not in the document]
MarahGamdou: > It is more readable to inline this.

I updated GetEvidences to make it more readable.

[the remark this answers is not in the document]
MarahGamdou: > Somehow I feel like srcmap is better, but up to you. I think a clear documentation is more important.

I will use `srcmap` if it's more clear.

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -366,28 +381,48 @@ func GetFuzzTargets(fuzzParameters *FuzzParameters, date string) ([]string, erro
...
+		return nil, err
+	}
+	evidences[0] = amber.ClaimEvidence{Role: role, URI: uri, Digest: *digest}
+
 	// Get the GCS absolute path of the file containing the coverage summary for the project on a given day.
-	evidences["project coverage"] = fmt.Sprintf("gs://%s/%s/reports/%s/linux/summary.json", CoverageBucket, fuzzParameters.ProjectName, date)
-	for _, fuzztarget := range fuzztargets {
-		// The role of the coverage evidence using the fuzztarget.
-		role := fmt.Sprintf("%s_%s_%v coverage", fuzzParameters.FuzzEngine, fuzzParameters.ProjectName, fuzztarget)
+	role = "project coverage"
```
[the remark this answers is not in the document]
MarahGamdou: I updated GetEvidences to make it more readable.

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: > I think all these could have the same role: "Fuzz target". What do you think?

I updated it to "fuzzTarget coverage" to have the same role for all fuzz-targets as discussed.

cmd/fuzzbinder/README.md
MarahGamdou: > Is the next step required too?

I agree, I did tests and the next step is not required. 

> Also, to complete the command above you have to follow a link to a browser and copy the token. Is not there a better way to authenticate?

I just had to connect via email in my tests. This is the command to run in [this documentation](https://googleapis.dev/python/google-api-core/latest/auth.html) for authentication. I linked this in this readme also for more details. However, I don't think that they mention something simpler.

cmd/fuzzbinder/README.md
MarahGamdou: Thanks! I updated it. However, I don't see why step two is optional. Unless the user does not want to test that they have access to the buckets before generating the claims they need gsutil, right?

cmd/fuzzbinder/main.go
```diff
@@ -13,13 +13,60 @@
...
+	flag.StringVar(&fuzzParameters.ProjectGitRepo, "git_repo", "",
+		"Required - GitHub repository of the project.")
+	// TODO(#175): Remove fuzzEngine and sanitizer from FuzzBinder inputs.
+	flag.StringVar(&fuzzParameters.FuzzEngine, "fuzzengine", "libFuzzer",
+		"Required - Fuzzing engine used for the project. Examples: libFuzzer, afl, honggfuzz, centipede.")
+	flag.StringVar(&fuzzParameters.Sanitizer, "sanitizer", "asan",
+		"Required - Fuzzing sanitizer used for the project. Examples: asan, ubsan, msan.")
+	flag.StringVar(&fuzzParameters.Date, "date", "",
+		"Required - Fuzzing date. The expected date format is YYYYMMDD. \n"+
+			"This date must not be more than 15 days prior to the date of execution of this cmd. \n"+
```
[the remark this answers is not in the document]
MarahGamdou: > Why? What happens if the date is too far in the past or even future?

In the future: 
You will get an error `could not generate the fuzzing claim: could not get the revision digest to generate the fuzzing claim: could not get oak/srcmap/20221220.json blob to extract revision hash: could not create a new reader for blob "oak/srcmap/20221220.json": storage: object doesn't exist`

In the past: 
You will get a FuzzClaim generated but with no fuzzing efforts and the "detectedCrashes" will be `false` (but it shouldn't be trusted because the log files are  deleted). You will also have the coverage data and it is reliable (because the coverage reports are kept for years).

internal/fuzzbinder/fuzzclaim.go
```diff
@@ -173,10 +172,136 @@ func parseFuzzClaimBytes(statementBytes []byte) (*intoto.Statement, error) {
...
 		return nil, err
 	}
 
 	return &statement, nil
 }
+
+// TODO(#171): Split generateFuzzClaimSpec into smaller functions.
+// generateFuzzClaimSpec generates a fuzzing claim specification using the
+// fuzzing reports of OSS-Fuzz.
+func generateFuzzClaimSpec(revisionDigest intoto.DigestSet, fuzzParameters *FuzzParameters, fuzzTargets []string) (*FuzzClaimSpec, error) {
```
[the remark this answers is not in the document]
MarahGamdou: > The solution is dependency inversion. Instead GetCoverage and other functions called below having to provide their resourced, the called function (e.g., generateFuzzClaimSpec or even better its callers) must provide the required resources to the downstream functions.

This part will be treated in the next PRs. 

>  I suggest you move this functionality to another file, or even another package. 

Done ! I moved it to another file.

cmd/fuzzbinder/main.go
```diff
@@ -13,13 +13,60 @@
...
+	flag.Parse()
+
+	// Get the absolute path for storing the fuzzing claim.
+	absFuzzClaimPath, err := filepath.Abs(fuzzClaimPath)
+	if err != nil {
+		log.Fatalf("could not get absolute path for storing the fuzzing claim: %v", err)
+	}
+
+	// Generate the fuzzing claim.
+	statement, err := fuzzbinder.GenerateFuzzClaim(fuzzParameters)
```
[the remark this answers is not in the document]
MarahGamdou: I agree. This part will be treated in the next PRs.

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: I agree but two returns are needed. One if `isGoodHash` and another in case it is not a good hash and we want zeros. I updated it to remove the variable declaration.

MarahGamdou:
Thanks for the reviews :)

cmd/fuzzbinder/main.go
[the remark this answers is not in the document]
MarahGamdou: > I suggest removing this from the help,
Done !

internal/fuzzbinder/fuzzscraper.go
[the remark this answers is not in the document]
rbehjati: The generated claim contains perTarget claims like the following:

```
                {
                    "name": "libFuzzer_oak_apply_policy",
                    "path": "oak/fuzz/fuzz_targets/apply_policy.rs",
                    "fuzzStats": {
                        "lineCoverage": "3.89% (215/5524)",
                        "branchCoverage": "0.00% (0/0)",
                        "detectedCrashes": false,
                        "fuzzTimeSeconds": 722232.9658203125,
                        "numberFuzzTests": 323023131
                    }
                },
```

I still don't understand where ` "path": "oak/fuzz/fuzz_targets/apply_policy.rs",` comes from. Is not it in some GCS file somewhere?

cmd/fuzzbinder/README.md
rbehjati: Yes. It is optional if it is installed from before.

cmd/fuzzbinder/main.go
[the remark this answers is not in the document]
MarahGamdou: I added an issue for this and added it as TODO in the code.

internal/fuzzbinder/fuzzscraper.go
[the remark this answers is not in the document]
MarahGamdou: In this version the path is fixed. It needs to be extracted from a coverage file in the future. I opened an issue and added it as a TODO.

rbehjati (approved):
Great job. Thanks :)

PR #178, opened 2022-12-19
Add Builder Image Digest

mariaschett:
Add Builder Image Digest in ProvenanceIR.

### DIFF
--- a/internal/verifier/verifier.go
+++ b/internal/verifier/verifier.go
@@ -154,6 +154,7 @@ func (verifier *ProvenanceMetadataVerifier) Verify() (VerificationResult, error)
 type ProvenanceIR struct {
 	BinarySHA256Digests []string
 	BuildCmds           [][]string
+	BuilderImageDigests []string
 }
 
 // FromSLSAv02 maps data from a validated SLSA v0.2 provenance to ProvenanceIR.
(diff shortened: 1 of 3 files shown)


internal/verifier/verifier.go
```diff
@@ -176,6 +177,12 @@ func FromAmber(provenance *amber.ValidatedProvenance) (ProvenanceIR, error) {
 	}
 	provenanceIR.BuildCmds = [][]string{buildCmd}
 
+	builderImageDigest, err := provenance.GetBuilderImageDigest()
+	if err != nil {
+		return provenanceIR, fmt.Errorf("could not convert from *amber.ValidatedProvenance: %v", err)
```
rbehjati, line 182:
```suggestion
		return provenanceIR, fmt.Errorf("could not get BuilderImageDigest from amber.ValidatedProvenance: %v", err)
```

internal/verifier/verifier.go
```diff
@@ -214,6 +221,16 @@ func (verifier *ProvenanceIRVerifier) Verify() (VerificationResult, error) {
...
 	}
 
+	// Verify BuilderImageDigest.
+	if verifier.Want.BuilderImageDigests != nil {
+		nextResult, err := verifier.Got.verifyBuilderImageDigest(verifier.Want)
+
+		if err != nil {
+			return combinedResult, fmt.Errorf("verify builder image digests failed: %v", err)
+		}
+		combinedResult.Combine(nextResult)
```
rbehjati, line 231: How about using https://pkg.go.dev/go.uber.org/multierr instead of `VerificationResult`? (in a separate PR)

pkg/amber/provenance.go
```diff
@@ -94,6 +95,20 @@ func (p *ValidatedProvenance) GetBuildCmd() ([]string, error) {
 	return buildConfig.Command, nil
 }
 
+// Guess the digest for the Builder Image.
```
rbehjati, line 98: Is this really a `guess`? 

The comments for exported functions should start with the function name `GetBuilderImageDigest`.

rbehjati (approved):
Nice. Thanks for the small PR :)

internal/verifier/verifier.go
mariaschett: Good idea: #184

internal/verifier/verifier.go
mariaschett: Done.

pkg/amber/provenance.go
mariaschett: Thanks! Done!

mariaschett (commented):
Done, thank you for the fast review!

PR #181, opened 2022-12-19
Encapsulate GCS connection into one struct to use it to apply dependency inversion 

MarahGamdou:
- Add gcsutil and update FuzzBinder accordingly.


cmd/fuzzbinder/main.go
```diff
@@ -50,8 +52,17 @@ func main() {
 		log.Fatalf("could not get absolute path for storing the fuzzing claim: %v", err)
 	}
 
+	// Get context
```
mariaschett, line 55: Which context are we getting and why? Otherwise I would drop the comment.

internal/gcsutil/util.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	defer reader.Close()
+	return reader, nil
+}
+
+// GetBlobData gets the data in a blob in a Google Cloud Storage bucket.
+func (client *GCSClient) GetBlobData(ctx context.Context, bucketName string, blobName string) ([]byte, error) {
+	reader, err := client.GetBlobReader(ctx, bucketName, blobName)
+	if err != nil {
+		return nil, err
```
mariaschett, line 98: I think we would also add context to the error here, e.g., "could not get blob reader: %v". Also below.

internal/gcsutil/util.go
```diff
@@ -0,0 +1,124 @@
...
+			logFiles = append(logFiles, blob)
+		}
+	}
+	if len(logFiles) == 0 {
+		return nil, fmt.Errorf("could not find log files in %s under %s", bucketName, relativePath)
+	}
+	return logFiles, nil
+}
+
+// GetBlobReader gets the file reader of a blob in a Google Cloud Storage bucket.
```
mariaschett, line 84: I am not sure extracting `GetBlobReader` is worth it. I think you only use it here, and it adds another indirection. What do you think?

internal/gcsutil/util.go
```diff
@@ -0,0 +1,124 @@
...
+	"google.golang.org/api/iterator"
+)
+
+// GCSClient contains a Google Cloud Storage client.
+type GCSClient struct {
+	Client *storage.Client
+}
+
+// NewClient creates and returns a new GCSClient.
+func NewClient(ctx context.Context) (*GCSClient, error) {
```
mariaschett, line 34: maybe `NewGCSClient`?

internal/gcsutil/util.go
```diff
@@ -0,0 +1,124 @@
...
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("could not create a new Google Cloud Storage client: %v", err)
+	}
+	defer client.Close()
```
mariaschett, line 39: Would we do that here? Would we not ask the owner of the client to close it? I am not sure. What do you think?

internal/gcsutil/util.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		blobs = append(blobs, attrs.Name)
+	}
+	return blobs, nil
+}
+
+// GetLogs returns all the log-files paths in a Google Cloud Storage bucket
+// under a given relative path.
+func (client *GCSClient) GetLogs(ctx context.Context, bucketName string, relativePath string) ([]string, error) {
+	blobs, err := client.ListBlobs(ctx, bucketName, relativePath)
```
mariaschett, line 68: If I understand correctly, we first list all the blobs and then we iterate over all the blobs again to process them. 

Does it make sense to do this in two steps? Can't we do it in one? 

Especially as we don't seem to need all the blobs anyway.

// EDIT: I just saw that we need ListBlobs somewhere else too. So you can ignore my comment or think about it :)

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -424,11 +318,12 @@ func GetFuzzTargets(fuzzParameters *FuzzParameters) ([]string, error) {
 }
 
 // addClaimEvidence adds an evidence to the list of the evidence files used by the fuzzscraper.
-func addClaimEvidence(evidences []amber.ClaimEvidence, blobName string, role string) ([]amber.ClaimEvidence, error) {
-	digest, err := getGCSFileDigest(CoverageBucket, blobName)
+func addClaimEvidence(ctx context.Context, client *gcsutil.GCSClient, evidences []amber.ClaimEvidence, blobName string, role string) ([]amber.ClaimEvidence, error) {
+	fileBytes, err := client.GetBlobData(ctx, CoverageBucket, blobName)
 	if err != nil {
-		return nil, fmt.Errorf("could not get digest of evidence %s: %v", blobName, err)
+		return nil, fmt.Errorf("could not get date in evidence file: %v", err)
```
mariaschett, line 324: Does this error make sense? do we try to get the data in evidence file here?

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -252,12 +196,7 @@ func getFuzzStatsFromScanner(lineScanner *bufio.Scanner) (*FuzzEffort, error) {
...
 
 // getFuzzEffortFromFile gets the fuzzingEffort from a single fuzzer log file.
-func getFuzzEffortFromFile(revisionDigest intoto.DigestSet, reader io.Reader) (*FuzzEffort, error) {
-	fileBytes, err := io.ReadAll(reader)
-	if err != nil {
-		return nil, fmt.Errorf(
-			"could not read data from log file to get fuzzing effort: %v", err)
-	}
+func getFuzzEffortFromFile(revisionDigest intoto.DigestSet, fileBytes []byte) (*FuzzEffort, error) {
 	isGoodHash, err := regexp.Match(revisionDigest["sha1"], fileBytes)
```
mariaschett, line 200: pre-existing: it seems that we want to do that check at least twice, e.g., in `crashDetectedInFile`. 

Does it make sense to either (i) have a function for that, such as `hasRevisionDigest`, or (ii) check before calling e.g., `getFuzzEffortFromFile`?

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -302,27 +236,12 @@ func crashDetected(revisionDigest intoto.DigestSet, reader io.Reader) (*Crash, e
...
-	if err != nil {
-		return nil, fmt.Errorf(
-			"could not get %s blob to compute SHA256: %v", blobName, err)
-	}
-	fileBytes, err := io.ReadAll(reader)
-	if err != nil {
-		return nil, fmt.Errorf(
-			"could not get data from %s reader to compute SHA256: %v", blobName, err)
-	}
+func getGCSFileDigest(fileBytes []byte) *intoto.DigestSet {
```
mariaschett, line 239: We are getting less errors in the function signatures---I think that is a good sign :) !

cmd/fuzzbinder/main.go
rbehjati: I agree, the comment does not add much. Also could you inline `context.Background()`?

cmd/fuzzbinder/main.go
```diff
@@ -50,8 +52,17 @@ func main() {
...
+	ctx := context.Background()
+
+	// Create new GCS client
+	client, err := gcsutil.NewClient(ctx)
```
rbehjati, line 59:
```suggestion
	client, err := gcsutil.NewClient(context.Background())
```

Does this work?

rbehjati, line 59: I suggest you wrap `ctx` in `Client`, instead of passing around as a separate argument.

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -197,31 +151,21 @@ func formatDate(fuzzParameters *FuzzParameters) string {
 	return hyphenDate
 }
 
-// getLogs gets the log-files list of a fuzz-target on a given day.
-func getLogs(fuzzParameters *FuzzParameters, fuzzTarget string) (*storage.BucketHandle, []string, error) {
+// getLogDirInfo gets the bucket and relative path of the directory
+// in which log-files are saved.
+func getLogDirInfo(fuzzParameters *FuzzParameters, fuzzTarget string) (string, string) {
```
rbehjati, line 156: Is it worth writing some unit tests for it?

internal/fuzzbinder/fuzzscraper.go
```diff
@@ -471,54 +366,43 @@ func GetEvidences(fuzzParameters *FuzzParameters, fuzzTargets []string) ([]amber
...
-					"could not analyze %s for crashes to GetCrashes: %v", blob, err)
-			}
-			if crash.Detected {
-				return crash, nil
-			}
+	for _, fileBytes := range listFileBytes {
+		crash, err := crashDetectedInFile(revisionDigest, fileBytes)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"could not analyze log data for crashes to GetCrashes: %v", err)
```
rbehjati, line 402:
```suggestion
				"could not analyze log data for crashes: %v", err)
```

internal/gcsutil/util.go
```diff
@@ -0,0 +1,124 @@
...
+	defer client.Close()
+	gcsClient := GCSClient{
+		Client: client,
+	}
+	return &gcsClient, nil
+}
+
+// ListBlobs returns all the objects paths in a Google Cloud Storage bucket
+// under a given relative path.
+func (client *GCSClient) ListBlobs(ctx context.Context, bucketName string, relativePath string) ([]string, error) {
```
rbehjati, line 48: Make `ctx` a field in `GCSClient` and replace `ctx` with `client.ctx` below. 

Also Golang style guide prefers shorter names: go/go-style/decisions#receiver-names. So perhaps replace `client` with `gc`.

internal/gcsutil/util.go
rbehjati: I agree. If it is only used in `GetBlobData` there is not much gain in separating it.

rbehjati (commented):
Nice refactoring. Thanks :) 

My main comment is about the context that is passed around. I could be wrapped as a field in GCSClient, then you won't have to pass it around separately.

cmd/fuzzbinder/main.go
MarahGamdou: I removed it

internal/gcsutil/util.go
MarahGamdou: Thanks for the suggestion ! I changed the naming.

internal/gcsutil/util.go
MarahGamdou: You mean close it in main? I think "It is recommended to close immediately after checking the error" that's why I kept it here.

internal/gcsutil/util.go
MarahGamdou: I agree. I removed it.

internal/gcsutil/util.go
MarahGamdou: Thanks, I agree. I updated it to avoid having two steps.

cmd/fuzzbinder/main.go
MarahGamdou: It is recommended by the Context package to not do so: 

`Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

func DoSomething(ctx context.Context, arg Arg) error {
    // ... use ctx ...
}`

However,

internal/gcsutil/util.go
MarahGamdou: I wonder if this is not considered a repetition in naming and if it is better to rename the `GCSClient` to `Client` instead, since they are all in the `gcsutil` package.

rbehjati: That is a good point. Since the package name is `gcsutil` I think you can drop `GCS`.

internal/gcsutil/util.go
mariaschett: Sounds good!

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: I added it.

internal/gcsutil/util.go
MarahGamdou: Done !

cmd/fuzzbinder/main.go
MarahGamdou: Yes, it works. I updated it.

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: Thanks !

internal/gcsutil/util.go
MarahGamdou: I renamed it to `c`.

MarahGamdou: > Make ctx a field in GCSClient and replace ctx with client.ctx below.

I replied above :)

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: I added a function for it.

internal/fuzzbinder/fuzzscraper.go
MarahGamdou: Oh, it's `data`, I fixed it. Thanks !

MarahGamdou:
Thanks for the reviews :)

internal/gcsutil/util.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// However, while it is generally important to pass the Context rather than
+// store it in another type, in the case below, this is not needed since there
+// is no risk of confusion. Indeed, the context is only used for the GCS client here.
+// Note that, if the use of context is extended in the future, then it should be
+// passed explicitly to each function that needs it as an argument.
+//
+// Client contains a Google Cloud Storage client and a context.Context.
+type Client struct {
+	StorageClient *storage.Client
```
rbehjati, line 46:
```suggestion
	storageClient *storage.Client
```

internal/gcsutil/util.go
```diff
@@ -0,0 +1,138 @@
...
+	Context       ContextInStruct
```
rbehjati, line 47:
```suggestion
	context       ContextInStruct
```
Make these fields private (unexposed), since I think no functions outside this package will ever need to access them directly.

internal/gcsutil/util.go
```diff
@@ -0,0 +1,138 @@
...
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+)
+
+// ContextInStruct contains contexts that can be used in
+// structures when there is no risk of confusion. Using
+// context.Context directly can lead to linting errors.
+type ContextInStruct context.Context
+
+// The documentation for context states:
```
rbehjati, line 33: Nice comment. Thanks :)

rbehjati (approved):
Very nice refactoring. Thanks.

internal/gcsutil/util.go
MarahGamdou: I agree :)

PR #182, opened 2022-12-19
Validate the fuzzing date in FuzzBinder

### LINKED ISSUE #176: Check the date range in FuzzBinder
Check the date range in the main before passing it to any downstream functions to make sure that it is no later than 15 days and not in the future.

MarahGamdou:
Fixes #176

### DIFF
--- a/cmd/fuzzbinder/main.go
+++ b/cmd/fuzzbinder/main.go
@@ -44,8 +44,8 @@ func main() {
 		"Optional - Output file name for storing the generated fuzzing claim.")
 	flag.Parse()
 
-	// validate fuzzing date.
-	err := fuzzbinder.ValidateDate(fuzzParameters.Date)
+	// Validate fuzzing date.
+	err := fuzzbinder.ValidateFuzzingDate(fuzzParameters.Date)
 	if err != nil {
 		log.Fatalf("could not validate the fuzzing date: %v", err)
 	}
--- a/cmd/fuzzbinder/main.go
+++ b/cmd/fuzzbinder/main.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/project-oak/transparent-release/internal/fuzzbinder"
 )
@@ -45,7 +46,9 @@ func main() {
 	flag.Parse()
 
 	// Validate fuzzing date.
-	err := fuzzbinder.ValidateFuzzingDate(fuzzParameters.Date)
+	// Current time in UTC time zone since it is used by OSS-Fuzz.
+	currentTime := time.Now().UTC()
+	err := fuzzbinder.ValidateFuzzingDate(fuzzParameters.Date, currentTime)
 	if err != nil {
 		log.Fatalf("could not validate the fuzzing date: %v", err)
 	}
--- a/cmd/fuzzbinder/main.go
+++ b/cmd/fuzzbinder/main.go
@@ -45,7 +45,6 @@ func main() {
 		"Optional - Output file name for storing the generated fuzzing claim.")
 	flag.Parse()
 
-	// Validate fuzzing date.
 	// Current time in UTC time zone since it is used by OSS-Fuzz.
 	currentTime := time.Now().UTC()
 	err := fuzzbinder.ValidateFuzzingDate(fuzzParameters.Date, currentTime)
--- a/cmd/fuzzbinder/main.go
+++ b/cmd/fuzzbinder/main.go
@@ -44,6 +44,12 @@ func main() {
 		"Optional - Output file name for storing the generated fuzzing claim.")
 	flag.Parse()
 
+	// validate fuzzing date.
+	err := fuzzbinder.ValidateDate(fuzzParameters.Date)
+	if err != nil {
+		log.Fatalf("could not validate the fuzzing date: %v", err)
+	}
+
 	// Get the absolute path for storing the fuzzing claim.
 	absFuzzClaimPath, err := filepath.Abs(*fuzzClaimPath)
 	if err != nil {
--- a/internal/fuzzbinder/util.go
+++ b/internal/fuzzbinder/util.go
@@ -1,0 +1,43 @@
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package fuzzbinder
+
+// This file provides utility functions for FuzzBinder tool.
+
+import (
+	"fmt"
+	"time"
+)
+
+// ValidateDate validates that the fuzzing date chosen to generate the fuzzing
+// claims is no more than 15 days prior to the date of execution of FuzzBinder cmd
+// and not in the future.
+func ValidateDate(date string) error {
+	claimDate, err := time.Parse("20060102", date)
+	if err != nil {
+		return fmt.Errorf(
+			"the format of %s is not valid: the date format should be YYYYMMDD.", date)
+	}
+	currentTime := time.Now()
+	duration := currentTime.Sub(claimDate).Hours() / 24
+	if duration > 15 {
+		return fmt.Errorf(
+			"the fuzzing logs on %s are deleted: select a more recent date.", date)
+	}
+	if duration < 0 {
+		return fmt.Errorf(
+			"no fuzzing logs generated for %s: do not select a date in the future.", date)
+	}
+	return nil
+}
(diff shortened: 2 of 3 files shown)


cmd/fuzzbinder/main.go
```diff
@@ -44,6 +44,12 @@ func main() {
 		"Optional - Output file name for storing the generated fuzzing claim.")
 	flag.Parse()
 
+	// validate fuzzing date.
```
mariaschett, line 47: nit: Capitalization, i.e., Validate

internal/fuzzbinder/util.go
```diff
@@ -0,0 +1,43 @@
...
+package fuzzbinder
+
+// This file provides utility functions for FuzzBinder tool.
+
+import (
+	"fmt"
+	"time"
+)
+
+// ValidateDate validates that the fuzzing date chosen to generate the fuzzing
```
mariaschett, line 23: What do you think of calling the func `ValidateFuzzingDate`---as it has functionality specific to fuzzing, i.e., the 15 days retention.

internal/fuzzbinder/util.go
```diff
@@ -0,0 +1,43 @@
...
+// claims is no more than 15 days prior to the date of execution of FuzzBinder cmd
+// and not in the future.
+func ValidateDate(date string) error {
+	claimDate, err := time.Parse("20060102", date)
```
mariaschett, line 27: Can you add a comment explaining the string "20060102"?

internal/fuzzbinder/util.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return fmt.Errorf(
+			"the format of %s is not valid: the date format should be yyyymmdd", date)
+	}
+	currentTime := time.Now()
+	duration := currentTime.Sub(claimDate).Hours() / 24
+	if duration > 15 {
```
mariaschett, line 34: Maybe we introduce a variable such as `ossFuzzLogRetentionDays := 15` and use it here by way of commenting?

internal/fuzzbinder/util.go
```diff
@@ -0,0 +1,43 @@
...
+// claims is no more than 15 days prior to the date of execution of FuzzBinder cmd
+// and not in the future.
+func ValidateDate(date string) error {
+	claimDate, err := time.Parse("20060102", date)
+	if err != nil {
+		return fmt.Errorf(
+			"the format of %s is not valid: the date format should be yyyymmdd", date)
+	}
+	currentTime := time.Now()
+	duration := currentTime.Sub(claimDate).Hours() / 24
```
mariaschett, line 33: I think that this computation does depend on when / where the tool is called.

Can we think of a better solution? When does a day start in the logs, i.e., which time zone? Can we go with that time zone?

However, maybe it is just sufficient to make sure that logs are actually---because that is what really matters to us? What do you think?

cmd/fuzzbinder/main.go
MarahGamdou: Thanks !

internal/fuzzbinder/util.go
MarahGamdou: I agree !

internal/fuzzbinder/util.go
MarahGamdou: I added a comment.

internal/fuzzbinder/util.go
MarahGamdou: Good idea !

internal/fuzzbinder/util.go
MarahGamdou: Another error will be returned if the logs are not found anyway (so this lack of precision will not cause a problem for FuzzBinder). However, I agree with the fact that this verification will depend on the time zone for the day -15 and current day.  

I will look at the exact time zone that is used for saving logs. However, I did not do it before because I am not sure if the  `ossFuzzLogRetentionDays`  is equal to 15.00 or about 15 (because it was an observation). For more precision, I will look if I can have access to both.

rbehjati: I agree, it would be more accurate to use the correct timezone. I am also fine with allowing 16 days and failing if there are any logs (I looked around a bit and it is not clear to me which time zone is used for the logs).

You can also shorten the code by using `IsAfter`/`IsBefore` and `AddDate(0, 0, -15)`.

internal/fuzzbinder/util.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// ValidateFuzzingDate validates that the fuzzing date chosen to generate the fuzzing
+// claims is no more than 15 days prior to the date of execution of FuzzBinder cmd
+// and not in the future.
+func ValidateFuzzingDate(date string) error {
```
rbehjati, line 26: You could add unit tests for it :)

rbehjati (approved):
Thanks for the small PR :)

internal/fuzzbinder/util.go
MarahGamdou: I think the time zone is UTC.

internal/fuzzbinder/util.go
MarahGamdou: I added  unit tests.

internal/fuzzbinder/util.go
MarahGamdou: I updated the time-zone and used `After/Before` to shorten the code.

MarahGamdou:
Thanks for the reviews :)

internal/fuzzbinder/util.go
```diff
@@ -0,0 +1,45 @@
...
+func ValidateFuzzingDate(date string, currentTime time.Time) error {
+	// The layout that represents the expected date format.
+	layout := "20060102"
+	fuzzClaimDate, err := time.Parse(layout, date)
+	if err != nil {
+		return fmt.Errorf(
+			"the format of %s is not valid: the date format should be yyyymmdd", date)
+	}
+	// The retention duration of the fuzzers logs saved in ClusterFuzz project bucket.
+	ossFuzzLogRetentionDays := 15
```
rbehjati, line 35: This could actually be a const declared at the top of the file.

internal/fuzzbinder/util.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// ValidateFuzzingDate validates that the fuzzing date chosen to generate the fuzzing
+// claims is no more than 15 days prior to the date of execution of FuzzBinder cmd
+// and not in the future.
+func ValidateFuzzingDate(date string, currentTime time.Time) error {
```
rbehjati, line 26:
```suggestion
func ValidateFuzzingDate(date string, referenceTime time.Time) error {
```

I am not sure `currentTime` is an accurate name.

internal/fuzzbinder/util_test.go
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package fuzzbinder
+
+import (
+	"testing"
+	"time"
+)
+
+func TestValidateFuzzingDate(t *testing.T) {
```
rbehjati, line 21: Nice test :)

internal/fuzzbinder/util_test.go
```diff
@@ -0,0 +1,44 @@
...
+		t.Fatalf("could not parse current time: %v", err)
+	}
+	validDates := []string{"20221206", "20221207", "20221208", "20221215",
+		"20221219", "20221220"}
+	invalidDates := []string{"20221321", "2023122", "20231221", "20221221",
+		"20221222", "20221223", "20221205", "20221204", "20221121"}
+	for _, date := range validDates {
+		err := ValidateFuzzingDate(date, currentTime)
+		if err != nil {
+			t.Errorf("unexpected fuzzing date validation error : got %v want nil", err)
```
rbehjati, line 35:
```suggestion
			t.Errorf("unexpected fuzzing date validation error: got %v want nil", err)
```

It is better to validate the actual text of the error. See for instance: https://github.com/project-oak/transparent-release/blob/a951a048581e468d6a426dded92594fda8058139/internal/endorser/endorser_test.go#L101-L104

internal/fuzzbinder/util_test.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+	for _, date := range invalidDates {
+		err := ValidateFuzzingDate(date, currentTime)
+		if err == nil {
+			t.Errorf("unexpected fuzzing date validation error : got nil want validation error")
```
rbehjati, line 41:
```suggestion
			t.Errorf("unexpected fuzzing date validation error: got nil want validation error")
```

Same as above.

internal/fuzzbinder/util.go
MarahGamdou: It is the time when the FuzzBinder cmd is executed. I changed it to `referenceTime` if it's more clear.

internal/fuzzbinder/util.go
MarahGamdou: I agree.

internal/fuzzbinder/util_test.go
MarahGamdou: Thanks ! I like the suggestion.

internal/fuzzbinder/util_test.go
MarahGamdou: Thanks ! I like the suggestion.

mariaschett (approved):
Great! Thank you! Also thank you for adding reference time and not making the tests fail in 15+ days :)

PR #183, opened 2022-12-19
Add validity period as input for FuzzBinder

### LINKED ISSUE #173: Add validity duration as an input parameter for the FuzzClaims
Add the claim validity duration as an input parameter for FuzzBinder.

MarahGamdou:
Fixes #173
- Add validity period: `notBefore` and `notAfter`
- Update how-to guide accordingly


cmd/fuzzbinder/main.go
```diff
@@ -42,6 +42,8 @@ func main() {
 		"Required - Fuzzing date. The expected date format is YYYYMMDD.")
 	fuzzClaimPath := flag.String("fuzzclaim_path", "fuzzclaim.json",
 		"Optional - Output file name for storing the generated fuzzing claim.")
+	validityDays := flag.Int("validity", 7,
+		"Optional - Validity period of the fuzzing claim that will be generated in days.")
```
mariaschett, line 46: Maybe add in the comment that the default is 7 days?

cmd/fuzzbinder/main.go
```diff
@@ -51,7 +53,7 @@ func main() {
 	}
 
 	// Generate the fuzzing claim.
-	statement, err := fuzzbinder.GenerateFuzzClaim(fuzzParameters)
+	statement, err := fuzzbinder.GenerateFuzzClaim(fuzzParameters, *validityDays)
```
mariaschett, line 56: I assume it does not make sense to add the "validity days" to the fuzzParameters?

mariaschett (approved):
Looks good, thanks!

cmd/fuzzbinder/main.go
MarahGamdou: I don't think so. The fuzzParameters are grouped together to be used for the fuzzscraper. The validityDays is used only once.

cmd/fuzzbinder/main.go
MarahGamdou: In this case I need to add the defaults to all the other comments for consistency. Do I do it in this PR?

mariaschett: No need to do this in this PR. Good idea, actually, to add comments for all in the same PR!

I do think that the command line providing the default values is nice to have. What do you think, @rbehjati ?

rbehjati: I agree that it is nice to have the defaults, but I'd check if the help already prints that. @MarahGamdou can you check and add the defaults only if the help does not already print them?

But I am not sure if 7 days is a reasonable default. As I described above, I think it is better to have flags for `notBefore` and `notAfter` and only make `notBefore` optional.

internal/fuzzbinder/fuzzgenerator.go
```diff
@@ -125,7 +125,7 @@ func GenerateFuzzClaim(fuzzParameters *FuzzParameters) (*intoto.Statement, error
 	currentTime := time.Now()
 	tomorrow := time.Now().AddDate(0, 0, 1)
 	// TODO(#173): Add validity duration as an input parameter.
-	endValidity := time.Now().AddDate(0, 0, 7)
+	endValidity := time.Now().AddDate(0, 0, validityDays)
```
rbehjati, line 128: Can validityDays be any number? Or is there a range limit?

cmd/fuzzbinder/main.go
```diff
@@ -42,6 +42,8 @@ func main() {
 		"Required - Fuzzing date. The expected date format is YYYYMMDD.")
 	fuzzClaimPath := flag.String("fuzzclaim_path", "fuzzclaim.json",
 		"Optional - Output file name for storing the generated fuzzing claim.")
+	validityDays := flag.Int("validity", 7,
```
rbehjati, line 45: I am not sure if number of days is the most convenient way of getting the validity range. How about directly getting `notBefore` and `notAfter` dates as input? We could use today's date as the default value for `notBefore`.

rbehjati (approved):
Nice. Thanks.

cmd/fuzzbinder/main.go
MarahGamdou: Yes, the help already adds the default. I checked it.

rbehjati: Thanks for checking :)

cmd/fuzzbinder/main.go
MarahGamdou: -  I agree that having `notAfter` and  `notBefore` as inputs is simpler for the user (not really for us, especially for `notBefore` ). 
Indeed, this can lead to some reliability problems: 
-  `notBefore` it should be verified to check it is set after the time of generating the claim to avoid tempering with the FuzzClaim. 
-  In the current version where `notBefore` is fixed, the exact time of claim generation can be found easily and reliably. This makes it easier to reproduce claims (since it means that only logs prior to that exact time were used). If the `notBefore` is an entry, this can create problems for reproducing claims, especially when logs are still being generated on a given day and FuzzBinder generates the fuzzClaim before the end of the day.
-  `notAfter` should be after `notBefore`

MarahGamdou: I suggest in that case to keep `notBefore` fixed by FuzzBinder and only allow `notAfter` to be an input.

MarahGamdou:
I think that we need to use the UTC time zone for consistency with OSS-Fuzz. What do you think ?

internal/fuzzbinder/fuzzgenerator.go
rbehjati: I meant if AddDate makes any assumptions about the number of days. I hope it does not. For instance, adding 40 days is the same as 1 month and 9-10 days.

cmd/fuzzbinder/main.go
rbehjati: Generally the contract for ClaimValidity is that `notBefore` is the same time or after the time the issue was generated. So I still think `notBefore` should be an input, but it has to be validated to be after the current time. 

For reproducibility, the date for which the claim was generated should be included in the FuzzClaimSpec. In particular, we should not use `fuzzParameters.Date` as the `notBefore` date.

rbehjati:
> I think that we need to use the UTC time zone for consistency with OSS-Fuzz. What do you think ?

Yes. Using UTC is generally a very good idea. Please use that for all date parameters.

cmd/fuzzbinder/main.go
MarahGamdou: Oh, there is already the `IssuedOn` for that. I forgot about it.

internal/fuzzbinder/fuzzgenerator.go
MarahGamdou: No, there is no assumption about the number of days from what I have seen.

cmd/fuzzbinder/main.go
rbehjati: Oh! Yes 👍🏼

MarahGamdou:
Thanks for the reviews :)

cmd/fuzzbinder/main.go
rbehjati: Just to be clear, `IssuedOn`, `NotBefore`, and `FuzzParameters.Date` could be the same date, but they don't have to be. The invariant that we should maintain is `FuzzParameters.Date <= IssuedOn <= NotBefore < NotAfter`.

MarahGamdou: I think that we have it now since 
- we validate that `CurrentTime = IssuedOn <= NotBefore < NotAfter` in the validity verification
- we validate `FuzzParameters.Date <= IssuedOn` when we validate the fuzzing date

What we have now: `<= IssuedOn -15 <= FuzzParameters.Date <= IssuedOn <= NotBefore < NotAfter`

MarahGamdou: The current problem is that it is hard to have `IssuedOn = NotBefore` since `NotBefore` comes from an input in `YYYYMMDD` format (without the exact time) while  `IssuedOn` is an exact time.

rbehjati: That is fine. We don't need to have `IssuedOn = NotBefore`. But if it happens in any way, it does not violate any rule.

PR #186, opened 2022-12-21
Use option pattern on  `ProvenanceIR`

mariaschett:
We use the option pattern to create a new `ProvenanceIR`. This makes it easy to extend the `ProvenanceIR` with new fields without changing the current implementations. A side benefit is this makes it easier for testing.

We also add Getters. We only need to set the fields once, so that makes it safer. Additionally, this simplifies the check on whether a field is present for the caller: an error indicates the field is not there.

We also move `ProvenanceIR` to `common` so we can later use it in the endorser.


internal/common/common.go
```diff
@@ -73,6 +73,87 @@ type ReferenceValues struct {
...
+// We use the ProvenanceIR to map different provenance formats.
+type ProvenanceIR struct {
+	binarySHA256Digest       string
+	buildCmd                 []string
+	builderImageSHA256Digest string
+}
+
+type ProvenanceIROption = func(p *ProvenanceIR)
+
+func NewProvenanceIR(binarySHA256Digest string, options ...func(*ProvenanceIR)) *ProvenanceIR {
```
rbehjati, line 86: Does it work?

```suggestion
func NewProvenanceIR(binarySHA256Digest string, options ... ProvenanceIROption) *ProvenanceIR {
```

mariaschett: Yes, it does! Now we can either keep the type `ProvenanceIROption` or replace it with` func(p *ProvenanceIR)` everywhere. I prefer the latter, but both are fine for me.

internal/common/common.go
```diff
@@ -73,6 +73,95 @@ type ReferenceValues struct {
...
+// We want to map different provenances of different build types to ProvenanceIR, so
+// all fields except for `binarySHA256Digest` are optional.
+type ProvenanceIR struct {
+	binarySHA256Digest       string
+	buildCmd                 []string
+	builderImageSHA256Digest string
+}
+
+// ProvenanceIROption is a helper type to enable the optional setting of fields in `NewProvenanceIR`
+type ProvenanceIROption = func(p *ProvenanceIR)
```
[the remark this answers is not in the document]
rbehjati: I don't see the need for it. Perhaps just remove it.

internal/common/common.go
```diff
@@ -73,6 +73,92 @@ type ReferenceValues struct {
...
+// We want to map different provenances of different build types to ProvenanceIR, so
+// all fields except for `binarySHA256Digest` are optional.
+type ProvenanceIR struct {
+	binarySHA256Digest       string
+	buildCmd                 []string
+	builderImageSHA256Digest string
+}
+
+// NewProvenanceIR creates a new proveance with given optional fields.
+// Every provenancy needs to have binary sha256 digest, so this is not optional.
```
rbehjati, line 86:
```suggestion
// Every provenancy needs to have a binary sha256 digest, so this is not optional.
```

internal/common/common.go
```diff
@@ -73,6 +73,92 @@ type ReferenceValues struct {
...
+
+// WithBuilderImageSHA256Digest adds a builder image sha256 digest when creating a new ProvenanceIR.
+func WithBuilderImageSHA256Digest(builderImageSHA256Digest string) func(p *ProvenanceIR) {
+	return func(p *ProvenanceIR) {
+		p.builderImageSHA256Digest = builderImageSHA256Digest
+	}
+}
+
+// GetBinarySHA256Digest gets the binary sha256 digest. Throws an error if the binary sha256 digest is empty.
+func (p *ProvenanceIR) GetBinarySHA256Digest() (string, error) {
```
rbehjati, line 110: Apparently the Golang convention is to drop `Get`. It is not followed everywhere in the code yet. Just thought I'd mention it if you are interested. But actually, it might be better to fix this in the entire codebase in a separate PR.

internal/common/common.go
```diff
@@ -73,6 +73,92 @@ type ReferenceValues struct {
...
+	if p.binarySHA256Digest == "" {
+		return "", fmt.Errorf("provenance does not have a binary SHA256 digest")
+	}
+	return p.binarySHA256Digest, nil
+}
+
+// GetBuildCmd gets the build cmd. Throws an error if the build cmd is empty.
```
rbehjati, line 117:
```suggestion
// GetBuildCmd gets the build cmd. Returns an error if the build cmd is empty.
```
nit: I think `return` better matches Golang terminology.

internal/common/common.go
```diff
@@ -73,6 +73,92 @@ type ReferenceValues struct {
...
+func (p *ProvenanceIR) GetBuildCmd() ([]string, error) {
+	if len(p.buildCmd) == 0 {
+		return nil, fmt.Errorf("provenance does not have a build cmd")
+	}
+	return p.buildCmd, nil
+}
+
+// GetBuilderImageSHA256Digest gets the builder image sha256 digest. Throws an error if the builder image sha256 digest is empty.
```
rbehjati, line 125:
```suggestion
// GetBuilderImageSHA256Digest gets the builder image sha256 digest. Returns an error if the builder image sha256 digest is empty.
```
nit!

internal/common/common.go
```diff
@@ -73,6 +73,92 @@ type ReferenceValues struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get build cmd from *amber.ValidatedProvenance: %v", err)
+	}
+
+	builderImageDigest, err := provenance.GetBuilderImageDigest()
+	if err != nil {
+		return nil, fmt.Errorf("could get builder image digest from *amber.ValidatedProvenance: %v", err)
+	}
+
+	provenanceIR := NewProvenanceIR(binarySHA256Digest,
```
rbehjati, line 148: I like this pattern :)

internal/verifier/verifier_test.go
```diff
@@ -182,10 +158,7 @@ func TestVerifyBuilderImageDigest_DigestFound(t *testing.T) {
 
 func TestVerifyBuilderImageDigest_DigestNotFound(t *testing.T) {
 	builderImageDigest := "9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9"
```
rbehjati, line 160: nit & pre-existing: 
```suggestion
	builderImageSHA256Digest := "9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9"
```

or perhaps follow the Golang guideline for shorter variable names in shorter scopes, and give it a cryptic short name :D

internal/verifier/verifier.go
```diff
@@ -245,55 +202,50 @@ func (got *ProvenanceIR) verifyBinarySHA256Digest(want common.ReferenceValues) (
...
 }
 
 // verifyBuilderImageDigest verifies that the given builder image digest matches a builder image digest in the reference values.
-func (got *ProvenanceIR) verifyBuilderImageDigest(want common.ReferenceValues) (VerificationResult, error) {
+func verifyBuilderImageDigest(want common.ReferenceValues, got *common.ProvenanceIR) (VerificationResult, error) {
 	result := NewVerificationResult()
 
-	if len(got.BuilderImageSHA256Digests) > 1 {
-		return result, fmt.Errorf("got multiple builder image digests (%s)", got.BuilderImageSHA256Digests)
+	gotBuilderImageDigest, err := got.GetBuilderImageSHA256Digest()
```
rbehjati, line 232: nit: 
```suggestion
	got, err := got.GetBuilderImageSHA256Digest()
```

internal/verifier/verifier.go
```diff
@@ -231,11 +186,13 @@ func (verifier *ProvenanceIRVerifier) Verify() (VerificationResult, error) {
 }
 
 // verifyBinarySHA256Digest verifies that the binary SHA256 in this provenance is contained in the given reference binary SHA256 digests (in want).
-func (got *ProvenanceIR) verifyBinarySHA256Digest(want common.ReferenceValues) (VerificationResult, error) {
+func verifyBinarySHA256Digest(want common.ReferenceValues, got *common.ProvenanceIR) (VerificationResult, error) {
 	result := NewVerificationResult()
 
-	if len(got.BinarySHA256Digests) != 1 {
-		return result, fmt.Errorf("got not exactly one actual binary SHA256 digest (%v)", got.BinarySHA256Digests)
+	gotBinarySHA256Digest, err := got.GetBinarySHA256Digest()
```
rbehjati, line 192: nit: 

```suggestion
	got, err := got.GetBinarySHA256Digest()
```

Since `want` is redeclared below, I wonder if it make sense to redeclare `got` too for better readability. (and to follow Golang's guideline for shorter variable names!)

rbehjati (approved):
I like this change! thanks :)
I left a few minor comments.

internal/common/common.go
mariaschett: Done.

internal/common/common.go
mariaschett: Sounds good!

internal/common/common.go
mariaschett: SG!

internal/common/common.go
mariaschett: SG!

internal/common/common.go
mariaschett: :)

internal/verifier/verifier.go
mariaschett: Ah, we cant do that because we use got. to declare `gotBinarySHA256Digest`. I renamed want to wantBinarySHA256Digest to make it more uniform. But not shorter variable names, I am afraid.

internal/verifier/verifier.go
mariaschett: As above.

internal/verifier/verifier_test.go
mariaschett: Done, thank you!

mariaschett (commented):
Thank you!

Issue #188, opened 2022-12-21
Use cmp in tests


rbehjati:
A conventional pattern in tests is to use the [cmp package](https://pkg.go.dev/github.com/google/go-cmp/cmp) and the following pattern:

```
if diff := cmp.Diff(got, want); diff != "" {
  t.Errorf("<descriptive error message referring to got and want>")
```

Ideally this should be followed in all tests and much as possible.

PR #189, opened 2022-12-21
Refactor Internal Endorser

### LINKED ISSUE #165: Generalize `endorser` to work on `ProvenanceIR`

mariaschett:
Fixes #165 .

This PR merges the `types.ValidatedProvenance` and `common.ProvenanceIR` into `types.ProvenanceIR`.

Benefits:
- only one internal, central representation of provenances
- reduces code duplication, in particular binary digest & binary name

Downside:
To avoid cyclic dependencies, we have a bit of a round-about.

1. To determine whether we have a SLSAv02 or Amber provenance, we need to first parse the intoto.statement with `types.ParseStatementData`. This only sets the field `provenanceStatement` in a ProvenanceIR `p`.
2. To determine the other fields in `ProvenanceIR` we need to call `common.SetProvenanceValues` on `p` (this has to live in common to avoid cyclic dependencies).

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -198,6 +198,7 @@ func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
 	// A *amber.ValidatedProvenance contains a SHA256 hash of a single subject.
 	binarySHA256Digest := provenance.GetBinarySHA256Digest()
 	buildType := amber.AmberBuildTypeV1
+	binaryName := provenance.GetBinaryName()
 
 	predicate, err := slsav02.ParseSLSAv02Predicate(provenance.GetProvenance().Predicate)
 	if err != nil {
@@ -219,6 +220,7 @@ func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
 
 	provenanceIR := NewProvenanceIR(binarySHA256Digest,
 		WithBuildType(buildType),
+		WithBinaryName(binaryName),
 		WithBuildCmd(buildCmd),
 		WithBuilderImageSHA256Digest(builderImageDigest),
 		WithRepoURIs(repoURIs))
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -76,103 +76,9 @@ type ReferenceValues struct {
 	RepoURI string `toml:"repo_uri"`
 }
 
-// ProvenanceIR is an internal intermediate representation of data from provenances.
-// We want to map different provenances of different build types to ProvenanceIR, so
-// all fields except for `binarySHA256Digest` are optional.
-type ProvenanceIR struct {
-	binarySHA256Digest       string
-	buildType                string
-	binaryName               string
-	buildCmd                 []string
-	builderImageSHA256Digest string
-	repoURIs                 []string
-}
-
-// NewProvenanceIR creates a new proveance with given optional fields.
-// Every provenancy needs to a have binary sha256 digest, so this is not optional.
-func NewProvenanceIR(binarySHA256Digest string, options ...func(p *ProvenanceIR)) *ProvenanceIR {
-	provenance := &ProvenanceIR{binarySHA256Digest: binarySHA256Digest}
-	for _, addOption := range options {
-		addOption(provenance)
-	}
-	return provenance
-}
-
-// WithBinaryName adds a binary name when creating a new ProvenanceIR.
-func WithBinaryName(binaryName string) func(p *ProvenanceIR) {
-	return func(p *ProvenanceIR) {
-		p.binaryName = binaryName
-	}
-}
-
-// WithBuildCmd adds a build cmd when creating a new ProvenanceIR.
-func WithBuildCmd(buildCmd []string) func(p *ProvenanceIR) {
-	return func(p *ProvenanceIR) {
-		p.buildCmd = buildCmd
-	}
-}
-
-// WithBuildType adds a build type when creating a new ProvenanceIR.
-func WithBuildType(buildType string) func(p *ProvenanceIR) {
-	return func(p *ProvenanceIR) {
-		p.buildType = buildType
-	}
-}
-
-// WithRepoURIs adds repo URIs referenced in the provenance when creating a new ProvenanceIR.
-func WithRepoURIs(repoURIs []string) func(p *ProvenanceIR) {
-	return func(p *ProvenanceIR) {
-		p.repoURIs = repoURIs
-	}
-}
-
-// WithBuilderImageSHA256Digest adds a builder image sha256 digest when creating a new ProvenanceIR.
-func WithBuilderImageSHA256Digest(builderImageSHA256Digest string) func(p *ProvenanceIR) {
-	return func(p *ProvenanceIR) {
-		p.builderImageSHA256Digest = builderImageSHA256Digest
-	}
-}
-
-// GetBinarySHA256Digest gets the binary sha256 digest. Returns an error if the binary sha256 digest is empty.
-func (p *ProvenanceIR) GetBinarySHA256Digest() (string, error) {
-	if p.binarySHA256Digest == "" {
-		return "", fmt.Errorf("provenance does not have a binary SHA256 digest")
-	}
-	return p.binarySHA256Digest, nil
-}
-
-// GetBinaryName gets the binary name. Returns an error if the binary name is empty.
-func (p *ProvenanceIR) GetBinaryName() (string, error) {
-	if p.binaryName == "" {
-		return "", fmt.Errorf("provenance does not have a binary name")
-	}
-	return p.binaryName, nil
-}
-
-// GetBuildCmd gets the build cmd. Returns an error if the build cmd is empty.
-func (p *ProvenanceIR) GetBuildCmd() ([]string, error) {
-	if len(p.buildCmd) == 0 {
-		return nil, fmt.Errorf("provenance does not have a build cmd")
-	}
-	return p.buildCmd, nil
-}
-
-// GetRepoURIs gets references to a repo in the provenance. There is no guarantee to get all the references to any repo.
-func (p *ProvenanceIR) GetRepoURIs() []string {
-	return p.repoURIs
-}
-
-// GetBuilderImageSHA256Digest gets the builder image sha256 digest. Returns an error if the builder image sha256 digest is empty.
-func (p *ProvenanceIR) GetBuilderImageSHA256Digest() (string, error) {
-	if p.builderImageSHA256Digest == "" {
-		return "", fmt.Errorf("provenance does not have a builder image SHA256 digest")
-	}
-	return p.builderImageSHA256Digest, nil
-}
-
 // FromProvenance validates and converts a provenance of arbitrary type to ProvenanceIR
 // TODO(#165): Remove types.ValidatedProvenance and perform the conversion directly on an intoto.statement.
-func FromProvenance(prov *types.ValidatedProvenance) (*ProvenanceIR, error) {
+func FromProvenance(prov *types.ValidatedProvenance) (*types.ProvenanceIR, error) {
 	predType := prov.PredicateType()
 	switch predType {
 	case intoto.SLSAV02PredicateType:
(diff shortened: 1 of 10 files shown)


internal/common/common.go
```diff
@@ -92,6 +93,13 @@ func NewProvenanceIR(binarySHA256Digest string, options ...func(p *ProvenanceIR)
 	return provenance
 }
 
+// WithBinaryName adds a binary name when creating a new ProvenanceIR.
+func WithBinaryName(binaryName string) func(p *ProvenanceIR) {
```
rbehjati, line 97: I am not sure if we want to verify this, but I could be wrong. Was there anything about this when you did the requirements analysis?

mariaschett: No, not at all, removed!

internal/common/common_test.go
```diff
@@ -151,25 +150,35 @@ func TestParseReferenceValues(t *testing.T) {
...
+	gotBuildCmd, err := got.GetBuildCmd()
 	if err != nil {
-		t.Fatalf("couldn't map provenance to ProvenanceIR: %v", err)
+		t.Fatalf("couldn't get build cmd")
 	}
+	testutil.AssertEq(t, "build cmd[0]", gotBuildCmd[0], "cp")
+	testutil.AssertEq(t, "build cmd[1]", gotBuildCmd[1], "testdata/static.txt")
+	testutil.AssertEq(t, "build cmd[2]", gotBuildCmd[2], "test.txt")
 
-	if diff := cmp.Diff(got, want, cmp.AllowUnexported(ProvenanceIR{})); diff != "" {
```
rbehjati, line 170: I had intentionally used `cmp.Diff`. This is the standard pattern that we should use everywhere (see #188). Please revert this change, and rewrite the test using `cmp`.

internal/common/common.go
```diff
@@ -76,161 +76,29 @@ type ReferenceValues struct {
...
-	return p.builderImageSHA256Digest, nil
-}
-
-// FromProvenance validates and converts a provenance of arbitrary type to ProvenanceIR
-// TODO(#165): Remove types.ValidatedProvenance and perform the conversion directly on an intoto.statement.
-func FromProvenance(prov *types.ValidatedProvenance) (*ProvenanceIR, error) {
-	predType := prov.PredicateType()
+// SetProvenanceData sets, depending on the predicate of the underlying provenance intoto.Statement,
+// the data for verifying the provenance.
+func SetProvenanceData(provenanceIR *types.ProvenanceIR) error {
```
rbehjati, line 81: I prefer the immutable style.

pkg/types/types.go
```diff
@@ -23,67 +23,136 @@ import (
 	"github.com/project-oak/transparent-release/pkg/intoto"
 )
 
-// ValidatedProvenance wraps an intoto.Statement representing a valid SLSA provenance statement.
-// A provenance statement is valid if it contains a single subject, with a SHA256 hash.
-type ValidatedProvenance struct {
+// ProvenanceIR wraps an intoto.Statement representing a valid SLSA provenance statement. A provenance statement is valid if it contains a single subject, with a SHA256 hash.
+// ProvenanceIR also holds internal intermediate representations of data from provenances. We want to map different provenances of different build types to ProvenanceIR, so
+// all fields except for `provenanceStatement` are optional.
+type ProvenanceIR struct {
```
rbehjati, line 29: I don't see why this is moved from `common` to here. In particular, `ProvenanceIR` is an internal representation, and belongs to `internal` not `pkg`, which by convention provides the public interface. 

Please move the content of this file to `common`, and leave only `ParseStatementData` in this file.

internal/verifier/verifier.go
```diff
@@ -124,35 +124,11 @@ func chdir(dir string) {
 	}
 }
 
-// ProvenanceMetadataVerifier verifies provenances by comparing the
-// content of the provenance predicate against a given set of expected values.
-type ProvenanceMetadataVerifier struct {
```
rbehjati, line 129: Why is the verifier removed?

pkg/amber/provenance.go
```diff
@@ -71,19 +71,59 @@ func validateSLSAProvenanceJSON(provenanceFile []byte) error {
...
 		return nil, err
 	}
 
-	return types.ParseStatementData(statementBytes)
+	provenanceIR, err := types.ParseStatementData(statementBytes)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse the provenance bytes: %v", err)
+	}
+
+	if err := SetAmberProvenanceData(provenanceIR); err != nil {
```
rbehjati, line 90: In general we want `ParseProvenanceFile` and `ParseStatementData` (or perhaps `ParseProvenanceData` considering the new change) to be only different in that one uses a file as input and the other uses a byte array as input. 

I suggest changing `SetAmberProvenanceData` to  take a byte array as input, first calling `types.ParseStatementData(statementBytes)` and then the rest. I also suggest renaming it to `ParseProvenanceData`. Then in `ParseProvenanceFile` as before, you only need to call one function (`ParseProvenanceData`) and no other logic.

pkg/types/types.go
```diff
@@ -23,67 +23,136 @@ import (
...
 	// The field is private so that invalid instances cannot be created.
-	provenance intoto.Statement
+	provenanceStatement      *intoto.Statement
```
rbehjati, line 31: Why do we still need to keep the provenance statement? It feels redundant.

pkg/intoto/slsa_provenance/v0.2/provenance.go
```diff
@@ -200,3 +201,20 @@ func GetMaterialsGitURI(pred ProvenancePredicate) []string {
 	}
 	return gitURIs
 }
+
+// SetSLSAv02ProvenanceData sets data to verify a SLSA v02 provenance in the given ProvenanceIR.
+func SetSLSAv02ProvenanceData(provenanceIR *types.ProvenanceIR) error {
```
rbehjati, line 206: Same as above, I suggest renaming this to `ParseProvenanceData`, and changing it to take a byte array as input.

pkg/amber/provenance.go
```diff
@@ -71,19 +71,59 @@ func validateSLSAProvenanceJSON(provenanceFile []byte) error {
...
+	}
+
+	return provenanceIR, nil
+}
+
+// SetAmberProvenanceData sets data to verify a Amber provenance in the given ProvenanceIR.
+func SetAmberProvenanceData(provenanceIR *types.ProvenanceIR) error {
+	buildType := AmberBuildTypeV1
+
+	predicate, err := slsav02.ParseSLSAv02Predicate(provenanceIR.GetProvenance().Predicate)
```
rbehjati, line 101: If you make the other changes I suggested, then here you could call `slsav02.ParseProvenanceData` and reuse more of the common functionality between `slsav02` and `amber`.

pkg/amber/provenance.go
mariaschett: I am not sure `SetAmberProvenanceData` will work with the dependencies. 

I was quite struggling to get those right, but I might be missing something. 

I suggest I'll first move the definition of `ProvenanceIR` from `types` to `common` as in the comment below and then revisit this comment.

mariaschett (commented):
Thanks for the review, I answered most comments.

Meta question: Are you happy with the change?

It is mainly a refactoring to get rid of the two definitions `ProvenanceIR` and `ValidatedProvenance`, but it is not strictly necessary.

I can drop it, if you prefer!

pkg/types/types.go
mariaschett: Update after discussion: We need the `intoto.Statement` ih the `ReproducibleProvenanceVerifier` for [LoadBuildConfig](https://github.com/project-oak/transparent-release/blob/a765fb136047cb4d6034012e3d91d76053705a70/internal/common/common.go#L129)

internal/common/common.go
rbehjati: Yes, and after a closer look at the code I think it is doable.

internal/common/common_test.go
rbehjati: Actually, what exactly is this test case testing? The name still says `TestFromProvenance_Amber`, but `FromProvenance` is replaced with `SetProvenanceData`, which is not even called in this test.

I think fixing this test will be easier once `SetProvenanceData` is fixed (ref my other comments on `SetXYZData`).

internal/verifier/verifier.go
rbehjati: Oh! I see. Sorry. I missed that its `Verify` is effectively only calling `Verify` on `ProvenanceIRVerifier`.

pkg/amber/provenance.go
rbehjati: Well yes, you don't want to introduce a dependency to `internal/common` in the files and packages in `pkg`. But if all these functions are tightly related to `ProvenanceIR` they should go into `common` too. 

But having a function `ParseProvenanceFile` in this file returning a data structure in `pkg` (e.g., `intoto.Statement`) gives a more convenient API.

Currently `types.ValidatedProvenance` is the glue between `internal` and `pkg`. The only thing that `types.ValidatedProvenance` adds to `intoto.Statement` is a validity check. If we only care about this validity in contexts where `ProvenanceIR` is used, then we can move the validation logic to a function that creates an instance of `ProvenanceIR`. Then we can remove `types.ValidatedProvenance`, and replace all its occurrences with `intoto.Statement`.

internal/common/common_test.go
```diff
@@ -151,25 +150,35 @@ func TestParseReferenceValues(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("couldn't get builder image digest")
 	}
+	testutil.AssertEq(t, "builder image digest", gotBuilderImageDigest, "9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9")
+
+	gotRepoURIs := got.GetRepoURIs()
+	testutil.AssertEq(t, "repo URI", gotRepoURIs[0], "https://github.com/project-oak/transparent-release")
 }
 
 func TestFromProvenance_Slsav1(t *testing.T) {
```
rbehjati, line 184: Update the name of the test.

pkg/amber/provenance.go
mariaschett: > Currently types.ValidatedProvenance is the glue between internal and pkg.

Yes, you are right, and I think types.ValidatedProvenance being the glue is the key. 

I just tried to move `ProvenanceIR` to internal.common, which brings me in a full circle of where I started: needing both: `ProvenanceIR` and `ValidatedProvenance`.

Now after all this, I think we should keep both: `ValidatedProvenance` and `ProvenanceIR`. 

The main reason is:

We have to go from bytes to `ValidatedProvenance` to `ProvenanceIR`. The last step, from `ValidatedProvenace` to `ProvenanceIR` depends on `ValidatedProvenance`. 

So I do not think it a good idea to merge the two types, if we want to keep the flexibility, especially as

> replace all its occurrences with `intoto.Statement`

Yes, that is what we would need to do, but I feel we are loosing information and control over the type in this case.

To summarize: I think we should keep `ValidatedProvenance` and `ProvenanceIR`. 

However, I am not sure about the guarantees the two types give (and related their names).

Should `ProvenanceIR` hold the binary digest/name? More generally: do we, internally, only want to work on `ProvenanceIR`?

rbehjati: > To summarize: I think we should keep `ValidatedProvenance` and `ProvenanceIR`.

Sounds good. 

> However, I am not sure about the guarantees the two types give (and related their names).
> 
> Should `ProvenanceIR` hold the binary digest/name? More generally: do we, internally, only want to work on `ProvenanceIR`?

I think it would be nice to do that. That would require converting `ValidatedProvenance` to `ProvenanceIR` as early as possible. Having separate fields in `ProvenanceIR` for digest and name makes sense to me, although IIRC, we won't have a reference value for name.

mariaschett:
Close in favour of the fresh PR #210 adding discussed changes.

PR #191, opened 2022-12-22
Add `buildType` to `ProvenanceIR`

### LINKED ISSUE #190: Add `buildType` to `ProvenanceIR`
The `buildType` indicates which kind of provenance this originates from. Right now, this could be Amber or SLSA.

This information is useful for identifying, which values we can expect in the ProvenanceIR to be present.

mariaschett:
Fixes #190


rbehjati (approved):
LGTM. Thanks.

PR #194, opened 2022-12-22
Remove duplicated types

rbehjati:
This PR unifies the two versions of `ValidatedProvenance` and moves the common part to a new package `types`. 

@mariaschett Let me know what you think.


pkg/types/types.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, fmt.Errorf("the provenance must have exactly one subject with a sha256 digest")
+	}
+
+	// statement.Predicate is now just a map, we have to parse it into an instance of slsa.ProvenancePredicate
+	predicateBytes, err := json.Marshal(statement.Predicate)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal Predicate map into JSON bytes: %v", err)
+	}
+
+	var predicate slsa.ProvenancePredicate
```
mariaschett, line 95: I think, the downside is: at this point you are locked into slsav02.ProvenancePredicate. 

How will this change when there is a slsa v1.0?

mariaschett:
Related to this the more general question: what do we guarantee for a `ValidatedProvenance`?

For know, I see the following guarantees:
- we can get a BinarySHA256Digest
- we can get a BinaryName.
- we can get a BuildType (maybe some guarantees on BuildType, too?).

I am also not sure, how `ValidatedProvenance` and `ProvenanceIR` relate to each other. I think they should.

Compare also #166 , where we were discussing a `ValidatedProvenanceValues`.

pkg/types/types.go
rbehjati: This should not change for slsa v1.0 I think. But we have to wait and see when the new format is out. 
Note that the files in `slsa_provenance/v02` are copied from the `slsa` repo, but we have full control on the structs. If needed, we can update these to a new common denominator.

rbehjati:
> Related to this the more general question: what do we guarantee for a `ValidatedProvenance`?
> 
> For know, I see the following guarantees:
> 
> * we can get a BinarySHA256Digest
> * we can get a BinaryName.
> * we can get a BuildType (maybe some guarantees on BuildType, too?).

I think that is about it. 

> I am also not sure, how `ValidatedProvenance` and `ProvenanceIR` relate to each other. I think they should.

We could potentially merge them into one. 

> Compare also #166 , where we were discussing a `ValidatedProvenanceValues`.

How does `ValidatedProvenanceValues` relate to `ProvenanceIR` now? Are not they going to be merged?

pkg/types/types.go
rbehjati: Here is the slsa v1 draf: https://github.com/slsa-framework/slsa/pull/525/files. Let's discuss this part separately.

mariaschett (approved):
As discussed, LGTM!

rbehjati:
> As discussed, LGTM!

Thanks. I made the changes we discussed :)

PR #196, opened 2022-12-23
Add check for Repo URIs

### LINKED ISSUE #192: Add verification step URI of repo
The project teams wants to verify whether the binary is built from resources from the project teams repository.

The URI of the repo is currently in: 

- `invocation -> configSource -> uri`, and in
- `materials -> uri`.
 
I suggest to 
-  add `repo_uri` as reference value, and 
- add `repo_uri_references` in `ProvenanceIR`, which contain references to the repo from the provenance. 

We assume, that the project team currently wants all references to point to the project team repo.

For reference: [slsa v0.2 example](https://slsa.dev/provenance/v0.2#example).

mariaschett:
Fixes #192 .


pkg/amber/provenance.go
```diff
@@ -100,6 +100,25 @@ func (p *ValidatedProvenance) GetBuildCmd() ([]string, error) {
 	return buildConfig.Command, nil
 }
 
+// GetConfigSourceURI returns the URI of the configuration.
+func (p *ValidatedProvenance) GetConfigSourceURI() string {
```
rbehjati, line 104: The same values can be retrieved the same way from slsav02 provenances too. Is that right?
Should not that we be done in this PR?

rbehjati: Now that the `ValidatedProvenance` types are unified, I think this can be moved to and be used both in `fromAmber` and `fromSlsav02`.

pkg/amber/provenance.go
```diff
@@ -100,6 +100,25 @@ func (p *ValidatedProvenance) GetBuildCmd() ([]string, error) {
...
+	return p.provenance.Predicate.(slsa.ProvenancePredicate).Invocation.ConfigSource.URI
+}
+
+// GetMaterialsGitURI returns references to a Git repo.
+func (p *ValidatedProvenance) GetMaterialsGitURI() []string {
+	materials := p.provenance.Predicate.([]slsa.ProvenanceMaterial)
```
rbehjati, line 110: Does this work? I think it should be changed to the following:
```suggestion
	materials := p.provenance.Predicate.Materials
```

Or better, change the signature to take a `slsa.ProvenancePredicate` as input. Similar to https://github.com/project-oak/transparent-release/blob/e6fbeb27dede4e91fa8553e38439f005c1639ccb/pkg/amber/provenance.go#L112.

pkg/amber/provenance.go
mariaschett: I don't think `GetConfigSourceURI` and `GetMaterialsGitURI` can be moved to `types`. 

We need to know which predicate it is, i.e., we would need to add a dependency on slsa in `types`. Or am I missing something?

pkg/amber/provenance.go
mariaschett: Done.

pkg/amber/provenance.go
rbehjati: Not to `types` but to `slsav02`. Because currently there is no `GetConfigSourceURI` for `slsav02`.

mariaschett: Do you think I should add `GetConfigSourceURI` and `GetMaterialsGitURI` here: https://github.com/project-oak/transparent-release/blob/main/pkg/intoto/slsa_provenance/v0.2/provenance.go ?

For me, I mainly need to be able to test, and as our current test provenance is amber, I added the implementations to the amber pkg.

internal/common/common.go
```diff
@@ -183,10 +198,15 @@ func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
 		return nil, fmt.Errorf("could get builder image digest from *amber.ValidatedProvenance: %v", err)
 	}
 
+	repoURIs := []string{}
+	repoURIs = append(repoURIs, amber.GetConfigSourceURI(predicate))
+	repoURIs = append(repoURIs, amber.GetMaterialsGitURI(predicate)...)
```
rbehjati, line 203: If the RepoURI appears both in `ConfigSource` and in `Materials` then `repoURIs` will contain two copies, right? That is probably ok, but can you add a comment explaining it?

internal/common/common.go
```diff
@@ -183,10 +198,15 @@ func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
...
+
 	provenanceIR := NewProvenanceIR(binarySHA256Digest,
 		WithBuildType(buildType),
 		WithBuildCmd(buildCmd),
-		WithBuilderImageSHA256Digest(builderImageDigest))
+		WithBuilderImageSHA256Digest(builderImageDigest),
+		WithRepoURIs(repoURIs))
```
rbehjati, line 209: I think this, and the code above should be added to `fromSLSAv02` too.

pkg/amber/provenance.go
```diff
@@ -70,6 +70,25 @@ func validateSLSAProvenanceJSON(provenanceFile []byte) error {
 	return nil
 }
 
+// GetConfigSourceURI returns the URI of the configuration.
+func GetConfigSourceURI(pred *slsav02.ProvenancePredicate) string {
```
rbehjati, line 74: Can you add tests for this and `GetMaterialsGitURI`? The tests in `verifier_test` do not cover these.

pkg/amber/provenance.go
rbehjati: Yes. https://github.com/project-oak/transparent-release/blob/main/pkg/intoto/slsa_provenance/v0.2/provenance.go  is a better place for those new functions. 

Are `GetConfigSourceURI` and `GetMaterialsGitURI` covered in any of the tests? I don't see them.

> and as our current test provenance is amber, I added the implementations to the amber pkg.

Do you mean we don't have any tests for `slsav02`? Since we have the code for it, we should have tests for it too. Actually, `slsav02` is more important than `amber`, as `amber` is deprecated.

pkg/amber/provenance.go
mariaschett: I would add tests in #198 (and modify `provenance.json`). This is then be covered by `TestGenerateEndorsement_SingleValidEndorsement`.

pkg/amber/provenance.go
mariaschett: I meant [`testdata/provenance.json`](https://github.com/project-oak/transparent-release/blob/main/testdata/provenance.json).

As far as I can tell, the tests rely a lot on `testdata/provenance.json`, but I don't think we can change this easily.

I am not sure how I could generate a `testdata/provenance.json` for slsa.v02.

Do you have an idea here?

internal/common/common.go
```diff
@@ -183,22 +198,42 @@ func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
...
+	// We collect repo uris from where they appear in the provenance to verify that they point to the same reference repo uri.
+	repoURIs := []string{}
+	repoURIs = append(repoURIs, slsav02.GetConfigSourceURI(*predicate))
+	repoURIs = append(repoURIs, slsav02.GetMaterialsGitURI(*predicate)...)
+
 	provenanceIR := NewProvenanceIR(binarySHA256Digest,
-		WithBuildType(buildType))
-	return provenanceIR
+		WithBuildType(buildType),
+		// TODO(#202): Add WithBuilderImageSHA256Digest.
```
rbehjati, line 233: I am not sure if there is a `BuilderImageSHA256Digest` for `slsav02.GenericSLSABuildType`, so you can remove the TODO.

internal/common/common.go
rbehjati: Thanks for adding the comment. I am still a bit uncomfortable with it. Also the comment does not explain the _why_, which is the most important thing you want from a comment. 

Do you expect the source URIs to be different or formatted differently in materials and in config source? I am asking because, a provenance should have exactly one source URI. 

Note that in the amber format we don't set the config source, and only set the Repo URI in materials. This is already the case in `[testdata/provenance.json](https://github.com/project-oak/transparent-release/blob/main/testdata/provenance.json).` So `GetConfigSourceURI` is not needed.

In generic slsa ([example](https://paste.googleplex.com/5755513575833600)), the same URI is included both in materials and in config source. So in this case too `GetConfigSourceURI` is not needed, because it is redundant.

pkg/amber/provenance.go
rbehjati: We need a new testdata file for slsav02 provenance. You can use [this example](https://paste.googleplex.com/5755513575833600). Perhaps rename `testdata/provenance.json` to `testdata/amber_provenance.json`.

pkg/amber/provenance.go
rbehjati: Great. Thanks.

pkg/intoto/slsa_provenance/v0.2/provenance.go
```diff
@@ -185,3 +186,22 @@ func ParseSLSAv02Predicate(predicate interface{}) (*ProvenancePredicate, error)
 
 	return &pp, nil
 }
+
+// GetConfigSourceURI returns the URI of the configuration.
+func GetConfigSourceURI(pred ProvenancePredicate) string {
```
rbehjati, line 191: Having looked at the provenance examples again, I'd suggest removing this entirely as it is either not used (in `amber`) or is redundant (in `slsav02`).

rbehjati (approved):
Thanks. LGTM, except that I see some unnecessary work for extracting RepoURI from ConfigSource that I suggest removing. Please see my comments.

pkg/amber/provenance.go
rbehjati: Actually, why are we using `testdata/provenance.json` and not `schema/provenance/v1/example.json`? Some tests use the latter.

mariaschett: Yes, I'll do that in #197 , too, I think there it fits.

mariaschett (commented):
Done, thank you!

pkg/amber/provenance.go
rbehjati: Thanks. I had a closer look, and I think perhaps `schema/provenance/v1/example.json` can be removed. But let's discuss that separately.

PR #203, opened 2022-12-29
Add test for FromProvenance

rbehjati:
Leftover from #194, which added the `FromProvenance` function.

### DIFF
--- a/internal/common/common_test.go
+++ b/internal/common/common_test.go
@@ -16,6 +16,7 @@ package common
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


internal/common/common_test.go
```diff
@@ -145,6 +146,24 @@ func TestParseReferenceValues(t *testing.T) {
 	testutil.AssertEq(t, "builder image digests[0]", referenceValues.BuilderImageSHA256Digests[0], "9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9")
 }
 
+func TestFromProvenance(t *testing.T) {
```
mariaschett, line 149: Should we add `TestFromProvenance_Amber`?

mariaschett (approved):
LGTM.

internal/common/common_test.go
rbehjati: Done.

rbehjati: Actually, renamed this one to `TestFromProvenance_Amber`. We can in the future add `TestFromProvenance_Slsav02`.

mariaschett: Renaming is what I meant :)

rbehjati: Added a (currently) negative test too, which uses a SLSA v1 provenance.

PR #204, opened 2022-12-29
Add Provenances SLSA v0.2 test

mariaschett:


internal/common/common_test.go
```diff
@@ -172,6 +173,32 @@ func TestFromProvenance_Amber(t *testing.T) {
...
+	statementBytes, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatalf("could not read the provenance file: %v", err)
+	}
+	provenance, err := types.ParseStatementData(statementBytes)
+	if err != nil {
+		t.Fatalf("couldn't parse the provenance file: %v", err)
+	}
+
+	// TODO(mschett): The current example for slsa_v02_provenance.json has a slightly different build type.
```
mariaschett, line 187: @rbehjati : 

The buildType currently defined in slsav02.GenericSLSABuildtype ends with `slsa-github-generator/generic@v1` and the one we have from our example provenance ends with `slsa-github-generator@v1`.

Can we fix the one in the slsa_v02_provenance.json manually? Or would this invalidate some digests?

mariaschett: I just fixed the build-type in `testdata/slsa_v02_provenance.json` to the one defined in [slsav02.GenericSLSABuildtype](https://github.com/project-oak/transparent-release/blob/main/pkg/intoto/slsa_provenance/v0.2/provenance.go#L35). 

I am a bit wary of manually fixing things in the example, I hope I have not messed anything up for future tests, but it does work for this one now.

testdata/slsa_v02_provenance.json
```diff
@@ -0,0 +1,261 @@
...
+        "invocation": {
+            "configSource": {
+                "uri": "git+https://github.com/project-oak/oak@refs/heads/main",
+                "digest": {
+                    "sha1": "1b128fb2556e4bdcc4f92552654bfbca9d2fb8c6"
+                },
+                "entryPoint": ".github/workflows/provenance.yaml"
+            },
+            "parameters": {},
+            "environment": {
```
rbehjati, line 26: For simplicity, I suggest removing the contents of this object. It is optional in the spec, and we don't have any checks on it anyway.

rbehjati (approved):
Nice. Thanks.

testdata/slsa_v02_provenance.json
mariaschett: Done, thanks!

PR #205, opened 2022-12-30
Add test where verification with reference values fails

mariaschett:
Adding a test where verification with reference fails.

### DIFF
--- a/internal/endorser/endorser_test.go
+++ b/internal/endorser/endorser_test.go
@@ -185,9 +185,7 @@ func TestLoadAndVerifyProvenances_NotVerified(t *testing.T) {
 		t.Fatalf("Could not load reference values: %v", err)
 	}
 
-	_, err = loadAndVerifyProvenances(referenceValues, []string{"file://" + tempPath1})
-
-	if err == nil || !strings.Contains(err.Error(), errorBinaryDigest) {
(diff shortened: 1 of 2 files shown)


internal/endorser/endorser_test.go
```diff
@@ -175,6 +174,34 @@ func TestLoadAndVerify_InconsistentNotVerified(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Could not load provenance: %v", err)
+	}
+
+	referenceValues, err := common.LoadReferenceValuesFromFile("../../testdata/different_reference_values.toml")
+	if err != nil {
+		t.Fatalf("Could not load reference values: %v", err)
+	}
+
+	_, err = loadAndVerifyProvenances(referenceValues, []string{"file://" + tempPath1})
```
rbehjati, line 188: This can be inlined I think:

```
if _, err = loadAndVerifyProvenances(referenceValues, []string{"file://" + tempPath1}); err != nil { ... }
```

rbehjati (approved):
Thanks :)

internal/endorser/endorser_test.go
mariaschett: Actually, we can't, I need `err` for the subsequent checks.

PR #208, opened 2023-01-05
Check whether value is present

### LINKED ISSUE #207: Verify only values from provenance
Currently we [`Verify` for every reference value](https://github.com/project-oak/transparent-release/blob/5083d96e94c2b0ffe91452f60200aca0ea42b2b2/internal/verifier/verifier.go#L162-L196)---even if the value does not occur in the provenance.

We only want to check those values we actually have in the current provenance.

We can do that, by adding for every field in `ProvenanceIR` a check on whether the value has been set, i.e., does not hold the default value any more.

For example: We add a `HasBuildCmd`, which is true if `ProvenanceIR` has been set.

NB: We can use `HasBuildCmd` also with `GetBuildCmd` [here](https://github.com/project-oak/transparent-release/blob/5083d96e94c2b0ffe91452f60200aca0ea42b2b2/internal/common/common.go#L138).

mariaschett:
Fixes #207 .

I also grouped the methods for each field together and moved them in order they appear in the struct.

- [x] Tests pass

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -105,19 +105,19 @@ func (p *ProvenanceIR) GetBinarySHA256Digest() (string, error) {
 	return p.binarySHA256Digest, nil
 }
 
-// WithBuildCmd adds a build cmd when creating a new ProvenanceIR.
+// WithBuildCmd sets the build cmd when creating a new ProvenanceIR.
 func WithBuildCmd(buildCmd []string) func(p *ProvenanceIR) {
 	return func(p *ProvenanceIR) {
 		p.buildCmd = &buildCmd
 	}
 }
 
-// HasBuildCmd returns true if the ProvenanceIR holds a build cmd.
+// HasBuildCmd returns true if the build cmd has been set in the ProvenanceIR.
 func (p *ProvenanceIR) HasBuildCmd() bool {
 	return p.buildCmd != nil
 }
 
-// GetBuildCmd gets the build cmd. Returns an error if the build cmd is empty.
+// GetBuildCmd gets the build cmd. Returns an error if the build cmd has not been set.
 func (p *ProvenanceIR) GetBuildCmd() ([]string, error) {
 	if !p.HasBuildCmd() {
 		return nil, fmt.Errorf("provenance does not have a build cmd")
@@ -125,19 +125,19 @@ func (p *ProvenanceIR) GetBuildCmd() ([]string, error) {
 	return *p.buildCmd, nil
 }
 
-// WithBuilderImageSHA256Digest adds a builder image sha256 digest when creating a new ProvenanceIR.
+// WithBuilderImageSHA256Digest sets the builder image sha256 digest when creating a new ProvenanceIR.
 func WithBuilderImageSHA256Digest(builderImageSHA256Digest string) func(p *ProvenanceIR) {
 	return func(p *ProvenanceIR) {
 		p.builderImageSHA256Digest = &builderImageSHA256Digest
 	}
 }
 
-// HasBuilderImageSHA256Digest returns true if the ProvenanceIR holds a builder image digest.
+// HasBuilderImageSHA256Digest returns true if the builder image digest has been set in the ProvenanceIR.
 func (p *ProvenanceIR) HasBuilderImageSHA256Digest() bool {
 	return p.builderImageSHA256Digest != nil
 }
 
-// GetBuilderImageSHA256Digest gets the builder image sha256 digest. Returns an error if the builder image sha256 digest is empty.
+// GetBuilderImageSHA256Digest gets the builder image sha256 digest. Returns an error if the builder image sha256 digest has not been set.
 func (p *ProvenanceIR) GetBuilderImageSHA256Digest() (string, error) {
 	if !p.HasBuilderImageSHA256Digest() {
 		return "", fmt.Errorf("provenance does not have a builder image SHA256 digest")
@@ -145,14 +145,14 @@ func (p *ProvenanceIR) GetBuilderImageSHA256Digest() (string, error) {
 	return *p.builderImageSHA256Digest, nil
 }
 
-// WithRepoURIs adds repo URIs referenced in the provenance when creating a new ProvenanceIR.
+// WithRepoURIs sets repo URIs referenced in the provenance when creating a new ProvenanceIR.
 func WithRepoURIs(repoURIs []string) func(p *ProvenanceIR) {
 	return func(p *ProvenanceIR) {
 		p.repoURIs = &repoURIs
 	}
 }
 
-// HasRepoURIs returns true if the ProvenanceIR holds repo URIs.
+// HasRepoURIs returns true if repo URIs have been set in the ProvenanceIR.
 func (p *ProvenanceIR) HasRepoURIs() bool {
 	return p.repoURIs != nil
 }
@@ -186,7 +186,7 @@ func FromProvenance(prov *types.ValidatedProvenance) (*ProvenanceIR, error) {
 }
 
 // fromAmber maps data from a validated Amber provenance to ProvenanceIR.
-// invariant: for every data X in a validated Amber provenance that can be mapped to a field in `ProvenanceIR`, `fromAmber` sets a value v for X by using `WithX(v)`.
+// invariant: for every data X in a validated Amber provenance that can be mapped to a field in `ProvenanceIR`, `fromAmber` sets a non-nil value v for X by using `WithX(v)`.
 func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
 	// A *amber.ValidatedProvenance contains a SHA256 hash of a single subject.
 	binarySHA256Digest := provenance.GetBinarySHA256Digest()
@@ -219,7 +219,7 @@ func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
 }
 
 // fromSLSAv02 maps data from a validated SLSA v0.2 provenance to ProvenanceIR.
-// invariant: for every data X in a validated SLSA v0.2 provenance that can be mapped to a field in `ProvenanceIR`, `fromSLSAv02` sets a value v for X by using `WithX(v)`.
+// invariant: for every data X in a validated SLSA v0.2 provenance that can be mapped to a field in `ProvenanceIR`, `fromSLSAv02` sets a non-nil value v for X by using `WithX(v)`.
 func fromSLSAv02(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
 	// A slsa.ValidatedProvenance contains a SHA256 hash of a single subject.
 	binarySHA256Digest := provenance.GetBinarySHA256Digest()
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -97,11 +97,12 @@ func NewProvenanceIR(binarySHA256Digest string, options ...func(p *ProvenanceIR)
 	return provenance
 }
 
-// WithBuildCmd adds a build cmd when creating a new ProvenanceIR.
-func WithBuildCmd(buildCmd []string) func(p *ProvenanceIR) {
-	return func(p *ProvenanceIR) {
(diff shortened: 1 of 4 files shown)


internal/common/common.go
```diff
@@ -111,49 +112,63 @@ func WithBuildType(buildType string) func(p *ProvenanceIR) {
...
-func (p *ProvenanceIR) GetRepoURIs() []string {
-	return p.repoURIs
+// WithBuilderImageSHA256Digest adds a builder image sha256 digest when creating a new ProvenanceIR.
+func WithBuilderImageSHA256Digest(builderImageSHA256Digest string) func(p *ProvenanceIR) {
+	return func(p *ProvenanceIR) {
+		p.builderImageSHA256Digest = builderImageSHA256Digest
+	}
+}
+
+// HasBuilderImageSHA256Digestis if the ProvenanceIR holds a builder image digest.
```
rbehjati, line 142:
```suggestion
// HasBuilderImageSHA256Digestis returns true if the ProvenanceIR holds a builder image digest.
```

internal/common/common.go
```diff
@@ -111,49 +112,63 @@ func WithBuildType(buildType string) func(p *ProvenanceIR) {
...
 }
 
+// WithRepoURIs adds repo URIs referenced in the provenance when creating a new ProvenanceIR.
+func WithRepoURIs(repoURIs []string) func(p *ProvenanceIR) {
+	return func(p *ProvenanceIR) {
+		p.repoURIs = repoURIs
+	}
+}
+
+// HasRepoURIs if the ProvenanceIR holds repo URIs.
```
rbehjati, line 162:
```suggestion
// HasRepoURIs returns true if the ProvenanceIR holds repo URIs.
```

internal/verifier/verifier.go
```diff
@@ -172,13 +172,13 @@ func (verifier *ProvenanceIRVerifier) Verify() (VerificationResult, error) {
 	}
 
 	// Verify HasBuildCmd.
-	if verifier.Want.WantBuildCmds {
+	if verifier.Got.HasBuildCmd() && verifier.Want.WantBuildCmds {
```
rbehjati, line 175: Just checking: "Does not have" is different from "cannot have". For instance, if the BuildType is Amber and the command is missing, we want this check to fail. Unless the conversion from Amber to ProvenanceIR checks that a command is present. Is that the case?

internal/verifier/verifier.go
```diff
@@ -172,13 +172,13 @@ func (verifier *ProvenanceIRVerifier) Verify() (VerificationResult, error) {
...
 		nextResult := verifyHasBuildCmd(verifier.Got)
 		combinedResult.Combine(nextResult)
 	}
 
 	// Verify BuilderImageDigest.
-	if verifier.Want.BuilderImageSHA256Digests != nil {
+	if verifier.Got.HasBuilderImageSHA256Digest() && verifier.Want.BuilderImageSHA256Digests != nil {
```
rbehjati, line 181: Same as above. I think we need to distinguish between "does not have" and "cannot have".

rbehjati (approved):
Thanks. The added test has revealed a corner case that got me thinking... should we allow a ProvenanceIR that does not have a BuildType?

mariaschett:
> Thanks. The added test has revealed a corner case that got me thinking... should we allow a ProvenanceIR that does not have a BuildType?

Hmm... by default I think optional is better with forwards compability. Do we have a reason why we don't want BuildType to be optional?

mariaschett:
> > Thanks. The added test has revealed a corner case that got me thinking... should we allow a ProvenanceIR that does not have a BuildType?
> 
> Hmm... by default I think optional is better with forwards compability. Do we have a reason why we don't want BuildType to be optional?

rbehjati:
> > Thanks. The added test has revealed a corner case that got me thinking... should we allow a ProvenanceIR that does not have a BuildType?
> 
> Hmm... by default I think optional is better with forwards compability. Do we have a reason why we don't want BuildType to be optional?
> 

Every provenance has a BuildType. So a valid ProvenanceIR must have a BuildType too. Otherwise how do we know which type of provenance statement it originated from?

> Taking this even further: Should we even make binary digest optional?

No. Every valid provenance must have at least one binary digest.

internal/verifier/verifier.go
mariaschett: You are absolutely right! 

So how I have achieved the distinction between "does not have" and "cannot have":

I have made every optional field in `ProvenanceIR` a pointer.

If `NewProvenance` is called `WithX`, then this pointer is set (possibly to the default value). 

`WithX` is called in `FromAmber` / `FromSlsav02`... So we have to maintain the invariant that `FromP` sets every value P _can_ have via `WithX`. 

Then `HasX` equals `false` (is `nil), this means that the provenance cannot have the value. 

If, e.g., for `build_cmd` we have an empty list (not `nil`), then the provenance does not have the value.

rbehjati: Ok, great. That sounds good. Thanks :)

internal/verifier/verifier_test.go
```diff
@@ -95,6 +95,30 @@ func TestReproducibleProvenanceVerifier_badCommand(t *testing.T) {
 	}
 }
 
+func TestVerifyHasNoValues(t *testing.T) {
+	// There are no values set apart from the binary digest.
+	got := common.NewProvenanceIR(binarySHA256Digest)
```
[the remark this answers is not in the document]
mariaschett: Not allowed anymore :)

internal/common/common.go
```diff
@@ -178,6 +186,7 @@ func FromProvenance(prov *types.ValidatedProvenance) (*ProvenanceIR, error) {
 }
 
 // fromAmber maps data from a validated Amber provenance to ProvenanceIR.
+// invariant: for every data X in a validated Amber provenance that can be mapped to a field in `ProvenanceIR`, `fromAmber` sets a value v for X by using `WithX(v)`.
```
rbehjati, line 189: Nice documentation :)

```suggestion
// invariant: for every data X in a validated Amber provenance that can be mapped to a field in `ProvenanceIR`, `fromAmber` sets a non-nil value v for X by using `WithX(v)`.
```

(or does "value" in Golang terminology imply non-nil?)

internal/verifier/verifier_test.go
```diff
@@ -184,8 +258,55 @@ func TestVerifyBuilderImageDigest_DigestNotFound(t *testing.T) {
 	}
 }
 
-func TestVerifyRepoURI_FoundURI(t *testing.T) {
-	got := common.NewProvenanceIR(binarySHA256Digest,
+func TestVerify_NeedsButHasEmptyBuilderImageDigest(t *testing.T) {
+	builderImageSHA256Digest := ""
+	got := common.NewProvenanceIR(binarySHA256Digest, amber.AmberBuildTypeV1, common.WithBuilderImageSHA256Digest(builderImageSHA256Digest))
```
rbehjati, line 263:
```suggestion
	got := common.NewProvenanceIR(binarySHA256Digest, amber.AmberBuildTypeV1, common.WithBuilderImageSHA256Digest(""))
```

And remove line 262.

internal/verifier/verifier_test.go
```diff
@@ -184,8 +258,55 @@ func TestVerifyBuilderImageDigest_DigestNotFound(t *testing.T) {
 	}
 }
 
-func TestVerifyRepoURI_FoundURI(t *testing.T) {
-	got := common.NewProvenanceIR(binarySHA256Digest,
+func TestVerify_NeedsButHasEmptyBuilderImageDigest(t *testing.T) {
+	builderImageSHA256Digest := ""
```
rbehjati, line 262: Is there a test where builderImage is not set (is nil), but a value is needed for it?
I expect the verification to pass in that case.

rbehjati (approved):
LGTM. Thanks :)
(I am a bit skeptical of using nil to have represent a conceptually valid situation, but let's proceed with it for now)

internal/common/common.go
```diff
@@ -78,80 +78,88 @@ type ReferenceValues struct {
...
+// HasBuildCmd returns true if the ProvenanceIR holds a build cmd.
+func (p *ProvenanceIR) HasBuildCmd() bool {
+	return p.buildCmd != nil
 }
 
-// WithRepoURIs adds repo URIs referenced in the provenance when creating a new ProvenanceIR.
-func WithRepoURIs(repoURIs []string) func(p *ProvenanceIR) {
-	return func(p *ProvenanceIR) {
-		p.repoURIs = repoURIs
+// GetBuildCmd gets the build cmd. Returns an error if the build cmd is empty.
```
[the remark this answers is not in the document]
mariaschett: Thank you, that is a good point. I rephrased to:

`Returns an error if the build cmd has not been set`. 

I think that captures the intention.

internal/common/common.go
```diff
@@ -78,80 +78,88 @@ type ReferenceValues struct {
...
-// GetBinarySHA256Digest gets the binary sha256 digest. Returns an error if the binary sha256 digest is empty.
-func (p *ProvenanceIR) GetBinarySHA256Digest() (string, error) {
-	if p.binarySHA256Digest == "" {
-		return "", fmt.Errorf("provenance does not have a binary SHA256 digest")
+// HasBuilderImageSHA256Digest returns true if the ProvenanceIR holds a builder image digest.
+func (p *ProvenanceIR) HasBuilderImageSHA256Digest() bool {
+	return p.builderImageSHA256Digest != nil
+}
+
+// GetBuilderImageSHA256Digest gets the builder image sha256 digest. Returns an error if the builder image sha256 digest is empty.
```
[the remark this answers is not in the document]
mariaschett: As above.

internal/common/common.go
mariaschett: Good point, added!

internal/common/common.go
```diff
@@ -211,6 +219,7 @@ func fromAmber(provenance *types.ValidatedProvenance) (*ProvenanceIR, error) {
 }
 
 // fromSLSAv02 maps data from a validated SLSA v0.2 provenance to ProvenanceIR.
+// invariant: for every data X in a validated SLSA v0.2 provenance that can be mapped to a field in `ProvenanceIR`, `fromSLSAv02` sets a value v for X by using `WithX(v)`.
```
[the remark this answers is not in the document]
mariaschett: Done.

internal/verifier/verifier_test.go
```diff
@@ -137,9 +211,9 @@ func TestVerifyHasBuildCmd_EmptyBuildCmds(t *testing.T) {
 	testutil.AssertEq(t, "no verification happened", result.IsVerified, true)
 }
 
-func TestVerifyBuilderImageDigest_DigestFound(t *testing.T) {
+func TestVerify_HasAndNeedsBuilderImageDigest(t *testing.T) {
 	builderImageSHA256Digest := "9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9"
```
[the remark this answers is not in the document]
mariaschett: Done.

internal/verifier/verifier_test.go
```diff
@@ -137,9 +211,9 @@ func TestVerifyHasBuildCmd_EmptyBuildCmds(t *testing.T) {
...
-	got := common.NewProvenanceIR(binarySHA256Digest, common.WithBuilderImageSHA256Digest(builderImageSHA256Digest))
+	got := common.NewProvenanceIR(binarySHA256Digest, amber.AmberBuildTypeV1, common.WithBuilderImageSHA256Digest(builderImageSHA256Digest))
```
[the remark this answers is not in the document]
mariaschett: Done.

internal/verifier/verifier_test.go
mariaschett: Done.

internal/verifier/verifier_test.go
mariaschett: This case is covered by `TestVerifyHasNoValues`.

mariaschett (commented):
Thank you very much for the re-review! :)

PR #212, opened 2023-01-10
Add Trusted Builder

### LINKED ISSUE #199: Verify Provenance is built by a Trusted Builder
We want to verify that a provenance is build by a Trusted Builder by checking against given reference values for the Trusted Builder.

In our example provenance we would be checking:
- [predicate -> builder](https://github.com/project-oak/transparent-release/blob/0df090b86f645b9253d5acf834020ca34386498d/testdata/provenance.json#L13) and maybe also
- [materials -> uri](https://github.com/project-oak/transparent-release/blob/0df090b86f645b9253d5acf834020ca34386498d/testdata/provenance.json#L30)

mariaschett:
Fixes #199 .

### DIFF
--- a/pkg/amber/provenance.go
+++ b/pkg/amber/provenance.go
@@ -110,8 +110,7 @@ func GetBuildCmd(predicate slsav02.ProvenancePredicate) ([]string, error) {
 
 // GetBuilder extracts and returns the builder from the given ProvenancePredicate.
 func GetBuilder(predicate slsav02.ProvenancePredicate) string {
-	// TODO(mschett): Return Builder.
-	return "Some Placeholder Builder"
+	return predicate.Builder.ID
 }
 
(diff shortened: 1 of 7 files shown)


pkg/amber/provenance.go
```diff
@@ -108,6 +108,11 @@ func GetBuildCmd(predicate slsav02.ProvenancePredicate) ([]string, error) {
 	return buildConfig.Command, nil
 }
 
+// GetBuilder extracts and returns the builder from the given ProvenancePredicate.
+func GetBuilder(predicate slsav02.ProvenancePredicate) string {
+	return predicate.Builder.ID
```
rbehjati, line 113: nit: Why is this wrapped in a function here, but not in the case of `SLSAv02`? Could not this be inlined where it is called?

rbehjati (approved):
Nice. Thanks :)

PR #213, opened 2023-01-12
Use multierr instead of VerificationResult

### LINKED ISSUE #184: Use `multierr` instead of `VerificationResult`
We could combine errors from different verification steps with [multierr](https://pkg.go.dev/go.uber.org/multierr#section-readme) instead of our home-cooked `VerificationResult`.

Downside: We can't distinguish between a verification step that failed because the verification failed (e.g., the provenance has an empty build cmd), and a failure in the process of verification.

However, the failure in the process are all due to `ProvenanceIR` having the wrong shape, which should go away after adding the changes discussed in #166 and #165 .

mariaschett:
Fixes #184 .

- [x] Tests pass

### DIFF
--- a/internal/verifier/verifier_test.go
+++ b/internal/verifier/verifier_test.go
@@ -62,10 +62,8 @@ func TestReproducibleProvenanceVerifier_invalidHash(t *testing.T) {
 		Provenance: provenance,
 	}
 
-	err = verifier.Verify()
-
 	wantErr := "failed to verify the digest of the built binary"
-	if err == nil || !strings.Contains(err.Error(), wantErr) {
+	if err = verifier.Verify(); err == nil || !strings.Contains(err.Error(), wantErr) {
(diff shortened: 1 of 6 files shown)


internal/verifier/verifier_test.go
```diff
@@ -115,12 +107,10 @@ func TestVerifyHasNoValues(t *testing.T) {
 	}
 
 	// We don't expect any verification to happen.
-	result, err := verifier.Verify()
+	err := verifier.Verify()
```
rbehjati, line 110: Can this be changed into `if err := ...; err != nil {...}`?

rbehjati (approved):
This is great. Thanks :)

internal/verifier/verifier_test.go
mariaschett: Done.

PR #216, opened 2023-01-25
Refactor endorser - load provenances before endorsement generation

### LINKED ISSUE #215: Change `GenerateEndorsement` to take provenances as ProvenanceIR
Currently [`GenerateEndorsement`](https://github.com/project-oak/transparent-release/blob/06ee7d94132967914fca88ba51928b54c21f2107/internal/endorser/endorser.go#L43) takes a list of URIs for provenances as input. This has to change to taking provenances as ProvenanceIRs. However, URIs (and provenance digests) are still needed as input as they are required as evidence in the generated endorsements. 

The following should happen before calling `GenerateEndorsement`:
- provenances should be fetched from the URIs
- any required preprocessing should be performed (e.g., if we get provenances as DSSE documents, we have to parse them, verify the signature, and parse the payload into a ProvenanceIR)
- wrap provenances as ProvenanceIR objects, provenance URIs, and digests in to a single object.

rbehjati:
Fixes #215

Adds a new struct `ParsedProvenance`, and refactors `GenerateEndorsement` to take an array of `ParsedProvenance`s instead of URIs. Tests are updated accordingly.

- [x] Tests pass

### DIFF
--- a/internal/endorser/endorser.go
+++ b/internal/endorser/endorser.go
@@ -17,8 +17,6 @@ package endorser
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


internal/endorser/endorser.go
```diff
@@ -32,58 +30,46 @@ import (
 	"github.com/project-oak/transparent-release/internal/verifier"
 	"github.com/project-oak/transparent-release/pkg/amber"
 	"github.com/project-oak/transparent-release/pkg/intoto"
-	"github.com/project-oak/transparent-release/pkg/types"
 )
 
+// ParsedProvenance contains contains a provenance in the internal
```
tiziano88, line 35:
```suggestion
// ParsedProvenance contains a provenance in the internal
```

internal/endorser/endorser.go
```diff
@@ -32,58 +30,46 @@ import (
...
+// ProvenanceIR format, and metadata about the source of the provenance. In
+// case of a provenance wrapped in a DSSE envelope, `SourceMetadata` contains
+// the URI and digest of the DSSE document, while `Provenance“ contains the
```
tiziano88, line 38:
```suggestion
// the URI and digest of the DSSE document, while `Provenance` contains the
```

tiziano88 (approved):

rbehjati:
Thanks for the review. Both fixed :)

Issue #187, opened 2022-12-21
Better follow go-style guide 


rbehjati:
Some style guidelines that we currently don't follow strictly:
- Drop `Get` in methods (see [Getters](http://go/go-style/decisions#getters))
- Short receiver names (See [Receiver names](http://go/go-style/decisions#receiver-names))

PR #219, opened 2023-01-27
Improve the code to better match Golang style & lints

rbehjati:
Ref #187

Additional changes were made to the verifier to reduce cyclomatic complexity.


rbehjati:
> LGTM.
> 
> nit: the PR does more than updating the function names, so you could reflect that in the title. However, you specify in the conversation and the tests pass :)

Thanks for the review. Updated the title to better reflect the changes.

PR #223, opened 2023-04-19
Remove builder

rbehjati:
Ref #145

Removing the builder, now that it is migrated to the [SLSA Framework](https://github.com/slsa-framework/slsa-github-generator) repo.

> It's a good idea to open an issue first for discussion.

- [x] Tests pass
- [x] Appropriate changes to README are included in PR

PR #225, opened 2023-04-20
Remove the Amber provenance format

### LINKED ISSUE #224: Deprecate Amber provenance
Now that 

- the builder is migrated to the [SLSA Framework](https://github.com/slsa-framework/slsa-github-generator), and removed from this repo (#223), and 
- the use of SLSA v0.2 is replaced by SLSA v1.0

the Amber provenance format can be deprecated.

This means that the verification and endorsement generation functionalities must be updated as well.

Additional cleanup might be possible in package common too.

rbehjati:
Fixes #224

- Removes Amber provenance files and format
- Removes BuildConfig and its related functionality (remaining from the builder that was removed in #223)
- Removes the `ReproducibleProvenanceVerifier`
- Updates tests, and testdata files to slsa-v0.2 format; essentially removing some reference values that should be reintroduced later, once we have support for SLSA v1.0

- [x] Tests pass


testdata/slsa_v02_provenance.json
```diff
@@ -12,20 +12,20 @@
     "predicate": {
         "builder": {
             "id": "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v1.2.0"
-        },
+                },
```
tiziano88, line 15: (nit) the indentation seems incorrect now?

tiziano88 (approved):

testdata/slsa_v02_provenance.json
rbehjati: Thanks for noticing this. Fixed.

rbehjati (commented):
Thanks for the quick review :)

PR #232, opened 2023-05-26
Extract commit hash from provenance

rbehjati:
While we don't verify the commit hash against a reference value, since ProvenanceIR is the internal representation, it should contain the commit hash.

This PR adds a new `commitSHA1Hash` field to `ProvenanceIR`. It also simplifies `RepoURIs` to be a single value. In all provenance formats, there is only one possible value for the source repo.

- [x] Tests pass

### DIFF
--- a/internal/model/provenance.go
+++ b/internal/model/provenance.go
@@ -207,7 +207,7 @@ func fromSLSAv02(provenance *ValidatedProvenance) (*ProvenanceIR, error) {
 		return nil, fmt.Errorf("could not parse provenance predicate: %v", err)
 	}
 
-	repoURI, commitHash := slsav02.GetGitURIAndDigest(*predicate)
+	repoURI, commitHash := slsav02.RepoURIAndDigest(*predicate)
 
 	// A ValidatedProvenance has a binary name.
 	binaryName := provenance.GetBinaryName()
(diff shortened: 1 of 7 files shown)


internal/model/provenance.go
```diff
@@ -42,7 +42,8 @@ type ProvenanceIR struct {
 	binaryName               string
 	buildCmd                 *[]string
 	builderImageSHA256Digest *string
-	repoURIs                 *[]string
+	repoURI                  *string
```
mariaschett, line 45: Just an idea:

From `slsav02.GetGitURIAndDigest`, I am wondering whether it makes sense to call this `gitURI` instead of `repoURI` and drop the concept of "repo" totally?

It seems in most places we are assuming Git (and Github?) already.

mariaschett (approved):
LGTM! Thanks!

mariaschett (approved):
LGTM! Thanks!

tiziano88 (approved):

internal/model/provenance.go
rbehjati: Good point on making the terminology consistent. I think `repoURI` is more suitable. Change all `GitURI`s to `RepoURI`.

rbehjati (commented):
Thanks for the reviews.

Note: I also changed turned functions on SLSA v1 and SLSA v0.2 into methods.

PR #235, opened 2023-06-08
Explain why DockerBasedBuildType is in draft

rbehjati:


pkg/intoto/slsa_provenance/v1/provenance.go
```diff
@@ -37,6 +37,8 @@ const (
 	PredicateSLSAProvenanceDraft = "https://slsa.dev/provenance/v1.0?draft"
 
 	// DockerBasedBuildType is the build type of container-based builds.
+	// The `draft` is there to signal that the format might need to change.
```
mariaschett, line 40: nit: maybe clarify: "is there" -> "in the URI signals", but it should be clear from context.

rbehjati: Thanks. This wording is better :)

rbehjati:
Thanks for the review.

Issue #234, opened 2023-06-01
Improve the use of strings in reference values & ProvenanceIR


rbehjati:
Currently, in the verifier, the endorser, and `ProvenanceIR`, we assign different meanings to a nil string and an empty ("") string. This is against some best practices:

> https://go.dev/doc/effective_go#allocation_new discusses zero values and "make the zero value useful".
> 
> [go-style/decisions#literal-zero-value-fields](http://go/go-style/decisions#literal-zero-value-fields) asserts the readability of zero-value construction.
> 
> [A counterargument](https://g3doc.corp.google.com/company/teams/protobuf/proposals/proto3-presence.md)

A more common pattern is to use a boolean to signal the presence or absence of a value. 

Ref: Readability review on cl/536413558

CC @mariaschett

PR #237, opened 2023-07-07
Replace ReferenceValues with proto and allow no provenance verification

### LINKED ISSUE #236: Allow provenance-less endorsement generation
Currently the endorser requires a provenance to be present when generating endorsement statements. 

We should relax that requirement, to allow other use cases, including endorsement generation for testing.

rbehjati:
Fixes #236 and #234

- [ ] Tests pass


mariaschett (commented):
Intermediate Comments :)

rbehjati (commented):
Thanks for the review.

proto/provenance_verification.proto
```diff
@@ -0,0 +1,45 @@
...
+
+option go_package = "proto/verification";
+
+message VerificationOptions {
+    // The absence of a reference provenance indicates that provenance verification can be skipped.
+    optional ProvenanceReferenceValues reference_provenance = 1;
+}
+
+message ProvenanceReferenceValues {
+    bool must_have_build_command = 1;
```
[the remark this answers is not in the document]
tiziano88: In proto3, everything is optional by default.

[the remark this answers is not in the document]
rbehjati: I removed optional from other fields too. [Here](https://github.com/project-oak/transparent-release/pull/237/commits/d3f2a6e158e2d1da64242643b77b44c0ee4f65a4) is the change in the generated Go code. Some oneof annotation and a oneof wrapper are removed. I am not sure if it has any serious impact on serialization.

internal/endorser/endorser.go
```diff
@@ -43,26 +44,35 @@ type ParsedProvenance struct {
...
-// must be provided. The endorsement statement is generated only if the provenance statements are
-// valid.
-func GenerateEndorsement(referenceValues *verification.ReferenceValues, validityDuration claims.ClaimValidity, provenances []ParsedProvenance) (*intoto.Statement, error) {
-	verifiedProvenances, err := verifyAndSummarizeProvenances(referenceValues, provenances)
+// GenerateEndorsement generates an endorsement statement for the given binary
+// and the given validity duration, using the given provenances as evidence and
+// VerificationOptions to verify them. If more than one provenance statements
+// are provided the endorsement statement is generated only if the provenance
+// statements are consistent and valid according to the input
+// VerificationOptions. If no provenances are provided, a provenance-less
```
mariaschett, line 52: A design question: might it happen, that I accidentially endorse something if I forget to add a provenance? Do we want to make this more explict? E.g., make `GenerateEndorsement` private but expose two methods `GenerateEndorsementWithProvenance` and `GenerateEndorsementWithoutProvenance`?

internal/endorser/endorser.go
```diff
@@ -43,26 +44,35 @@ type ParsedProvenance struct {
...
+// endorsement is generated, if the input VerificationOptions does not contain
+// a reference provenance.
+func GenerateEndorsement(binaryName, binaryDigest string, verOpt *prover.VerificationOptions, validityDuration claims.ClaimValidity, provenances []ParsedProvenance) (*intoto.Statement, error) {
+	if (verOpt.GetSkipProvenanceVerification() == nil) && (verOpt.GetReferenceProvenance() == nil) {
```
mariaschett, line 56: Just checking: Don't you get from `oneof` in the proto definition that exactly one of `verOpt` is not nil?

internal/endorser/endorser.go
```diff
@@ -73,22 +83,35 @@ func verifyAndSummarizeProvenances(referenceValues *verification.ReferenceValues
 		provenancesData = append(provenancesData, p.SourceMetadata)
 	}
 
-	errs := multierr.Append(verifyConsistency(provenanceIRs), verifyProvenances(referenceValues, provenanceIRs))
+	var errs error
+	if len(provenanceIRs) > 0 {
```
mariaschett, line 87: I am trying to understand what happens if we have multiple provenances but `SkipProvenanceVerification`.

IIUC the control flow lands here, but then `GetReferenceProvenance` would be nil? Is this okay with `verifyProvenances`?

On a higher level: What do we expect to happen then?

mariaschett (commented):
I might be missing something on how `SkipProvenanceVerification` and the list of provenances interact. Could you have a look and let me know?

golangci-linters.yaml
```diff
@@ -65,7 +65,7 @@ linters:
     - noctx
     - nolintlint
     - nonamedreturns
-    - nosnakecase
+    # - nosnakecase
```
tiziano88, line 68: Remove?

pkg/claims/endorsement_test.go
```diff
@@ -68,6 +68,25 @@ func TestNotAfterBeforeNotBeforeEndorsement(t *testing.T) {
 	}
 }
 
+func TestGenerateProvenanceLessEndorsement(t *testing.T) {
+	newNotBefore := time.Now().AddDate(0, 0, 1)
```
tiziano88, line 72: Ideally the tests (and the functions under tests) would not rely on the current time, but would accept a reference time from the outside. This will make them easier to test (without having to depend on relative times), and also avoid actual bugs (TOCTOU and related issues, if multiple functions get their own `time.Now()` independently at slightly different times).

proto/provenance_verification.proto
```diff
@@ -0,0 +1,50 @@
...
+    Digests reference_builder_image_digests = 3;
+    string reference_repo_uri = 4;
+    StringAllowList reference_builders = 5;
+}
+
+// An allow list of digests, represented as a mapping from cryptographic hash
+// function names, to their allow listed values. 
+message Digests {
+    // Keys are cryptographic hash algorithms (e.g., sha256).
```
tiziano88, line 42: I thought we were using the numeric multihash id somewhere else (from https://github.com/multiformats/multicodec/blob/master/table.csv ).

in particular `sha256` is ambiguous; it should be `sha2-256` or something like that, to be unambiguos -- but the numeric id solves that problem (without having to actual implement full multihash -- I don't think we should do that for now).

tiziano88 (approved):

docs/development-guidelines.md
```diff
@@ -23,3 +23,27 @@ You need to have:
 - Format files: `./scripts/formatting.sh`
 - Check linting: `./scripts/linting.sh`
 - Additional checks: `go vet ./...`
+
+## Using protocol buffers
```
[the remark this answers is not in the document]
rbehjati: Thanks. We used to use bazel in this repo, and it caused a lot of problems. I think this is fine for now.

golangci-linters.yaml
rbehjati: The template kept these as commented out. I think this is for convenience to be able to enable or disable them easily.

internal/endorser/endorser_test.go
```diff
@@ -22,17 +22,19 @@ import (
...
 	"github.com/project-oak/transparent-release/pkg/claims"
+	prover "github.com/project-oak/transparent-release/pkg/proto/verification"
+	"google.golang.org/protobuf/encoding/prototext"
 )
 
 const (
-	binaryHash        = "d059c38cea82047ad316a1c6c6fbd13ecf7a0abdcc375463920bd25bf5c142cc"
-	binaryName        = "oak_functions_freestanding_bin"
-	errorBinaryDigest = "do not contain the actual binary SHA256 digest"
+	binaryHash                   = "d059c38cea82047ad316a1c6c6fbd13ecf7a0abdcc375463920bd25bf5c142cc"
```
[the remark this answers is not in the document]
rbehjati: Done.

pkg/claims/endorsement_test.go
rbehjati: For the endorsement to be valid the notBefore time should be after the `issued` time, which uses the current time. For the test to pass, we have to use the current time here.

proto/provenance_verification.proto
rbehjati: Not in this repo. This repo follows the in-toto and SLSA style of using digests. We can change that in the future, but that is out of the scope of this PR.

internal/endorser/endorser.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/project-oak/transparent-release/internal/verification"
 	"github.com/project-oak/transparent-release/pkg/claims"
 	"github.com/project-oak/transparent-release/pkg/intoto"
+	prover "github.com/project-oak/transparent-release/pkg/proto/verification"
```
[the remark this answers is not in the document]
rbehjati: I agree, but I cannot think of a better name. I don't like long names with underscore in them. 
Do you have any suggestions?

rbehjati (commented):
Thanks for the reviews.

internal/endorser/endorser.go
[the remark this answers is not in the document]
rbehjati: Renamed the other `verification` into `verifier`. The only files in it now are `verifier.go`  and `verifier_test.go` so I think that makes sense.

mariaschett (approved):
Looks good, added a couple more comments, where I wasn't 100% sure.

PR #242, opened 2023-08-03
Add a terminating newline at the end of generated endorsement statement

rbehjati:


cmd/endorser/main.go
```diff
@@ -108,6 +108,9 @@ func main() {
 		log.Fatalf("Failed marshalling the endorsement: %v", err)
 	}
 
+	// Add a newline at the end of the file.
+	newline := []byte("\n")[0]
```
tiziano88, line 112: maybe

```suggestion
	newline := '\n'
```

andrisaar, line 112: Could you do something like `var newline byte = '\n'`? Going through a rune (and a slice) feels inelegant.

tiziano88 (approved):

andrisaar (approved):

cmd/endorser/main.go
rbehjati: I tried, but it does not work. Here is the error: `cannot use newline (variable of type rune) as byte value in argument to append`. And I just learnt that `rune` is a 32 bit type. Not sure how to convert it to byte. Changed it to `newline := byte('\n')`.

rbehjati: Changed it to `newline := byte('\n')`.

PR #243, opened 2023-09-01
Includes SHA2-384 and SHA2-512 digests in the endorsement statement

rbehjati:
To be able to compute the additional digests, the endorser is updated to take as input the path to the binary, so that it can load the binary bytes and compute additional digests.

Example subject included in the endorsement:

```json
    "subject": [
        {
            "name": "stage0_bin",
            "digest": {
                "sha2-256": "70a4fae8cd01e8e509f0d29efe9cf810192ad9b606fcf66fb6c4cbfee40fd951",
                "sha2-384": "d408084265af6776b269dc126c2d51129472b89b85de3bd9cd558d8bd617ae34dedd68f1a0954bd1c7f9b81a943dde81",
                "sha2-512": "e66d3449c2ee6e1a066e5dbae690b025eacd63aa0e969ba8cf97ed93940ec0634ce7f54816e0b80f0e84d9c3e493395da11145b6f639a3a3527f52ccb8aa6db3"
            }
        }
    ],
```


rbehjati:
@conradgrobler This should allow supporting SHA384 measurements for stage0 (ref https://github.com/project-oak/oak/pull/4269#discussion_r1310539443)

@tiziano88 I've explicitly included both SHA2-384 and SHA2-512 in the digests, although SHA2-384 is a truncated version of SHA2-512. The explicit inclusion of the digests makes the verification logic easier. Also, IIUC, SHA2-384 is preferred over SHA2-512, as it is not vulnerable to a length extension attack. So I think we'd prefer SHA2-384 over  SHA2-512.

@thmsbinder Heads up... as a result of this change, the API for the endorser is changed.

cmd/endorser/main.go
```diff
@@ -147,18 +145,21 @@ func parseDateOrDefault(date string, value time.Time) (time.Time, error) {
...
-	digest := digest{
-		alg:   parts[0],
-		value: parts[1],
+
+	sum256 := sha256.Sum256(bytes)
+	sum512 := sha512.Sum512(bytes)
+	sum384 := sha512.Sum384(bytes)
+
+	digestSet := intoto.DigestSet{
+		"sha256":   hex.EncodeToString(sum256[:]),
```
conradgrobler, line 160: Shouldn't this be "sha2-256" for consistency? Or does some other process still assume we will use "sha256"?

conradgrobler (approved):
Thanks.

cmd/endorser/main.go
rbehjati: I kept `sha256` for consistency with SLSA provenances, but the endorsement verification is under our control, and for that we can use `sha2-256`. Changed to `sha2-256`.

rbehjati (commented):
Thanks for the quick review.

internal/endorser/endorser_test.go
```diff
@@ -103,7 +105,8 @@ func TestGenerateEndorsement_SingleProvenance_EndorseProvenanceLess(t *testing.T
 	}
 	provenances := createProvenanceList(t, []string{"../../testdata/slsa_v02_provenance.json"})
 
-	statement, err := GenerateEndorsement(binaryName, binaryDigestSha256, verOpts, createClaimValidity(7), provenances)
+	digests := map[string]string{"sha256": binaryDigestSha256}
```
tiziano88, line 108: Should this also be `sha2-256` (and below)?

internal/model/validation.go
```diff
@@ -30,13 +30,31 @@ type sigstoreBundle struct {
 	DSSEEnvelope *dsse.Envelope `json:"dsseEnvelope"`
 }
 
-// ValidatedProvenance wraps an intoto.Statement representing a valid SLSA provenance statement.
-// A provenance statement is valid if it contains a single subject, with a SHA256 hash.
+// ValidatedProvenance wraps an intoto.Statement representing a valid SLSA
+// provenance statement. A provenance statement is valid if it contains a
+// single subject, with a SHA256 hash.
```
tiziano88, line 35:
```suggestion
// single subject, with a SHA2-256 digest.
```

tiziano88 (approved):

thmsbinder (approved):

internal/model/validation.go
rbehjati: Thanks. Done.

internal/endorser/endorser_test.go
rbehjati: Good catch. Thanks. Fixed.

rbehjati (commented):
Thanks for the review.

